Issue #24, opened 2015-03-13
Multiple logging handlers may be instatiated with platform.agent.utils.setup_logging()


ChrisFreeman:
For a given agent with multiple modules, it's not always obvious where to call to `utils.setup_logging()`. Alternatively, calling it in many modules has the undesired establishing multiple logging handlers that, in turn, cause multiple log message to appear.

As opposed to the "don't do that" advice, it seems better to add a check to `utils.setup_logging()` to check for existing handler before instantiating and adding another one to the root logger.

PR #25, opened 2015-03-13
Prevent setup_logging multiple handlers

ChrisFreeman:
Resolves issue #24 
Updated volttron.platform.agent.utils.set_logging() to check for
existing root logging handler before instantiating new handler.


hashstat:
I verified the issue and the patch. Thank you for your contribution.

PR #26, opened 2015-03-24
Allow port selection for modbus tcp and default to 502 if port undefined

devmull:
Specifying a modbus tcp port can be useful in cases where only non-privileged ports are available. For instance, simulating a modbus slave using pyModSlave on port 502 requires root whereas ports above 1024 (such as 8585) do not.


devmull:
Oops, I didn't have to modify modbus.py after all!  Please ignore.  Will submit another PR with just the comment in driver.ini.

Issue #54, opened 2015-05-11
Import problem in main.py when using restricted code


jhaack:
Fails due: PasswordRequiredException, SSHException, RSAKey

Add to imports

craig8:
This is because my service username has expired.  I will get that fixed today.

jhaack:
OBE

PR #55, opened 2015-05-11
Addresses missing imports for #54

jhaack:
Fixes missing imports


hashstat:
The paramiko import was intentionally removed knowing it would break the restricted components until they can be updated to use VIP. Rather than import modules we removed from the requirements, I suggest fixing the restricted components to work with the new protocols. I'm recommend we don't merge this PR.

jhaack:
Works for me

craig8:
Won't be merged.

PR #82, opened 2015-06-08
Migrate ControlService and PubSubService to new vipagent package

timothykang:
`PlatformAgent` had to be updated along the way to deal with strong connection method assumptions in `control._send_agent` (i.e. presence of RPC and channel subsystem methods).

Would be great if @hashstat and @craig8 could review.

### DIFF
--- a/Agents/PlatformAgent/platformagent/agent.py
+++ b/Agents/PlatformAgent/platformagent/agent.py
@@ -62,11 +62,9 @@ import requests
 import os
 import os.path as p
 import re
-import shutil
-import tempfile
 import uuid
 
-from volttron.platform import vip, jsonrpc, control
+from volttron.platform import vip, jsonrpc
 from volttron.platform.control import Connection
 from volttron.platform.agent import utils
 from volttron.platform.agent.vipagent import (BaseAgent, periodic, onevent,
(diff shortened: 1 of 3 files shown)


hashstat:
Thanks for taking a crack at this, @timothykang. I was currently working on making similar changes to channel.py and main.py. There were some additional changes required for the agent transfer and it exposed some issues with the latest channel subsystem implementation. I think we should hold off on merging this PR. I should have my changes in place today.

hashstat:
The PR is made up of two commits. Commit a3dc5c409810d02ef8275d308997dbbf3a0eb8d7 can be ignored as several commits I recently pushed cover the changes. The other commit cannot be used as is and I would recommend that, if possible, the agent continue to call _send_agent() so it can take advantage of the changes there and any future updates. I'll defer that decision to @craig8.

timothykang:
Thanks for the review and the updates. I'm glad to see that my proposed changes were inline with how the new package was intended to be used.

At the moment, agent installation by `PlatformAgent` is broken by the changes to `control._send_agent` in 4854403, because `_send_agent` calls methods provided by the new package, while `PlatformAgent` still inherits from the deprecated `agent.vipagent` module.

I'll can update this pull request to migrate `PlatformAgent` to the new package, maintaining that tight coupling between it and `_send_agent`, though in my opinion there's a little of code smell in doing so. Perhaps there's a way to update _send_agent so that it's more reusable across agent implementations.

hashstat:
The intent is to move most (if not all) of _send_agent() into a more generic send_file(). But they will both assume an instance of volttron.platform.vipagent.Agent as the first argument. This is a largely a transitional issue and @craig8 knew it was coming down the pipe.

Please do update the PR with _send_agent() intact. Thanks.

timothykang:
Since the changes in this pull request are obsolete, I'll close it. I've created #85 to capture the need to fix agent installation by `PlatformAgent`.

Issue #66, opened 2015-05-19
Implement Unregistering a platform from VOLTTRON Central


jhaack:
Be able to unregister a platform from VOLTTRON Central making it available for management by another instance of VOLTTRON Central

craig8:
@jhaack  says it's all good

Issue #80, opened 2015-06-05
Platforms should announce when they come online


craig8:
The platform.agent should announce that it is available or unavailable through the /platform pubsub topic.

Issue #77, opened 2015-05-28
Expose historian through VC


craig8:
Modify platform agent to publish to the VC historian.

craig8:
Historian is exposed through the base_historian.py file.   Implementations of SQLHistorianAgent provide incite into how this is done.

Issue #70, opened 2015-05-21
Platforms do not re-register if VOLTTRON Central restarts


jhaack:
If VOLTTRON Central is restarted after a platform has registered, this registration is lost.

VC should store registration information in order to resume state. Alternatively/additionally, platforms could periodically re-register with VC.

PR #84, opened 2015-06-08
Several issues handled

craig8:
Fixed issue #66, #70, #77, #80 

These are available from the platform manger not necessarily all the way out to the user as yet.  @jhaack and @hashstat would probably be the best to look over this especially @jhaack putting this out on highroad.


Agents/PlatformAgent/platformagent/agent.py
```diff
@@ -193,39 +244,54 @@ def route_request(self, id, method, params):
...
+        @Core.receiver('onstart')
+        def starting(self, sender, **kwargs):
+            self.vip.pubsub.publish(peer='pubsub', topic='/platform',
+                                    message='available')
+        @Core.receiver('onstop')
+        def stoping(self, sender, **kwargs):
+            self.vip.pubsub.publish(peer='pubsub', topic='/platform',
+                                    message='leaving')
+
+#         @Core.receiver('onsetup')
```
craig8: The comments here should be removed.

PR #121, opened 2015-07-20
Demo comms merge

jhaack:


craig8:
This will be done via the sqlhistorian merge that will be submitted shortly.

Issue #243, opened 2015-12-02
Concrete SQLHistorians should be case insensitive


craig8:
This will prevent case miss spellings to occur.  SMAP is case sensitive in its topic structure, however SQL by default is case insensitive as far as data is conserned.

All of the db functs modules in the db directory of the historian should be modified for case insensitive topics.

craig8:
Looks like we need to add some things to make this work.  Add a place to lowercase the topic_name before stored in the historian topic_map.  While keeping the actual case in the topic_name that is inserted/updated in the database.

Issue #163, opened 2015-08-26
Add topics for Analysis and Record in base_historian


jhaack:
Analysis will mimic devices and should be timestamp/values that can be graphed. Where devices is based on /all analysis will be point by point. This is to make it simpler for people to publish out points without a lot of formatting.

- [x] Record is a black box with no expectation of being graphed or understood by anything else. Applications can use this to record state for debugging later by the developer.

- [x] Document on wiki page

PR #245, opened 2015-12-02
Feature/historian

craig8:
Fixes tasks #224, #243 and #163

[Review on Reviewable](https://reviewable.io/reviews/volttron/volttron/245)

### DIFF
--- a/services/core/SQLHistorian/sqlhistorian/db/mysqlfuncts.py
+++ b/services/core/SQLHistorian/sqlhistorian/db/mysqlfuncts.py
@@ -143,7 +143,7 @@ class MySqlFuncts(DbDriver):
         return '''REPLACE INTO data values(%s, %s, %s)'''
         
     def insert_topic_query(self):
-        return '''INSERT INTO topics (topic_name) values (%s)'''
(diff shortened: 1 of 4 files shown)


craig8:
@jhaack please review.

My testing involved modifying an sqlite historian agent and a listener agent publishing to /request.  The following code snippet could be placed in the heartbeat function of the listener to duplicate this.

```
self.vip.pubsub.publish(
            'pubsub', 'record/listeneragent', headers, {'foo':'wolrd'}).get(timeout=5)
        result = self.vip.rpc.call(
                                           #Send this message to the platform historian
                                           #Using the reserved ID
                                           'platform.historian', 
                                           #Call the query method on this agent
                                           'query', 
                                           #query takes the keyword arguments of:
                                           #topic, then optional: start, end, count, order
#                                            start= "2015-10-14T20:51:56",
                                           topic='record/listeneragent',
                                           count = 20,
                                           #RPC uses gevent and we must call .get(timeout=10)
                                           #to make it fetch the result and tell 
                                           #us if there is an error
                                           order = "LAST_TO_FIRST").get(timeout=10)
        print('Query Result', result)
```

PR #277, opened 2016-01-08
Feature/testing bdd

schandrika:
[Review on Reviewable](https://reviewable.io/reviews/volttron/volttron/277)


craig8:
@schandrika I did a merge on my local box (mint) and found that the version number for the revision is not the same as it is in red hat.  On Ubuntu based system there is an integer dash ubnutu so 27-yubuntu as the end.  Nice huh!

craig8:
Actually it's even worse its really the string 5.6.27-0ubuntu0.14.04.1 as the version number.

PR #289, opened 2016-01-12
Add basic failover capability to the master driver.

mikeroup:
[Review on Reviewable](https://reviewable.io/reviews/volttron/volttron/289)


kmonson:
You should put the delay in the MasterDriver before it starts up the sub drivers similar to how the staggering is done.

kmonson:
I forgot about the per driver interval. @mikeroup did it correctly the first time.

Issue #217, opened 2015-10-28
Actuator 3.0 cannot interact with 2.0 agents


jhaack:
The Actuator in 3.0 cannot interact with 2.0 agents. This should be documented with an example of how to transition agents to talk to the 3.0 Actuator. Alternatively, enhance the actuator to work with legacy applications.

craig8:
Can we get a status on this @kmonson

kmonson:
We should test it again. During Actuator testing we discovered the
techniques used to deal with 2.0 agents was incorrect. We fixed it to use
the utility functions for dealing with 2.0 compatibility but have not had
an opportunity to test it.

PR #384, opened 2016-03-08
Feature/test actuator

schandrika:
Done with whatever actuator test cases I intended to write. @kmonson I think you need to review only the changes done for issue #217

[Review on Reviewable](https://reviewable.io/reviews/volttron/volttron/384)


kmonson:
I'll fix the merge problems.

PR #408, opened 2016-03-17
Update mongo to behave more like sqlhistorians

craig8:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/408)


schandrika:
Done. Should I delete the branch now?

PR #451, opened 2016-04-12
Resolves #71, 'volttron-ctl remove' fails to remove agent.

### LINKED ISSUE #71: Agent isn't removed when removing running agent
When an agent is runing (has a pid in status window) if the user removes the agent via volttron-ctl remove uuid...the agent still shows in the volttron-ctl status as being a running process.  

Trying to stop the agent at this point produces an error not found (i.e. volttron-ctl stop uuid).  volttron-ctrl clear does not remove the agent from volttron-ctl status list.

Restarting volttron does remove the agent from the volttron-ctl status list.

ffreckle:
The problem occurred because volttron-ctl was not making an rpc call to remove
agent but was calling the AIPPlatform instance directly.  The inline call removed the
installed agent files but did not remove the agent from volttron's list of agents.

This commit includes a new test in test_platformwrapper: test_can_remove_agent,  along
with a corresponding change in PlatformWrapper.remove_agent which now makes a remote
call rather than calling the aip inline.  Its possible this was a workaround for
some other problem but I don't see any new tests failing because of the change.  Let me know if I've missed something.

I do see an indeterminate number of 'auth' tests failing (6-9) before and after
this change which I assume is related to my environment.  All other tests are passing.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/451)

### DIFF
--- a/volttron/platform/control.py
+++ b/volttron/platform/control.py
@@ -318,7 +318,7 @@ def remove_agent(opts):
             return 10
         for agent in match:
             _stdout.write('Removing {} {}\n'.format(agent.uuid, agent.name))
-            opts.aip.remove_agent(agent.uuid)
+            opts.connection.call('remove_agent', agent.uuid)
 
 def _calc_min_uuid_length(agents):
     n = 0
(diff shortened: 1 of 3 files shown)


craig8:
Thanks @ffreckle !

PR #575, opened 2016-05-24
Take out debug messages which reprint entire publishes

jhaack:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/575)

### DIFF
--- a/volttron/platform/agent/base_historian.py
+++ b/volttron/platform/agent/base_historian.py
@@ -368,7 +368,6 @@ class BaseHistorian(Agent):
         return output_topic
 
     def _capture_record_data(self, peer, sender, bus, topic, headers, message):
-#         _log.debug('Capture record data {}'.format(message))
         # Anon the topic if necessary.
         topic = self._get_topic(topic)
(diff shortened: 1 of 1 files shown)


craig8:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[volttron/platform/agent/base_historian.py, line 371 \[r1\]](https://reviewable.io:443/reviews/volttron/volttron/575#-KIZnqk1V80pivyd08Ii:-KIZnqk1V80pivyd08Ij:387436821) ([raw file](https://github.com/volttron/volttron/blob/808c51ab453015b5373e8acb0d2ffa58a634ec9c/volttron/platform/agent/base_historian.py#L371)):*
> ```Python
> 
>     def _capture_record_data(self, peer, sender, bus, topic, headers, message):
> #         _log.debug('Capture record data {}'.format(message))
> ```

Why are we commenting them out rather than removing them?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/volttron/volttron/575)*

Issue #673, opened 2016-07-28
MasterWebService adds 'allow-all' authentication entry


zbeech:

craig8:
ahh yeah.  relic from where we were doing this.

jhaack:
@craig8  @zbeech  Still an issue?

zbeech:
It is technically still in the code: https://github.com/VOLTTRON/volttron/blob/develop/volttron/platform/web.py#L202

We can remove this, and it won't affect the behavior because the AuthFile will always have at least one entry (the platform itself). Thus, `if not entries` will always be false.

zbeech:
merged into develop

PR #674, opened 2016-07-28
Remove unconditional addition of "allow all" entry

zbeech:
This does not fully address the issue (#673), but this is a small step. Prior to this commit, MasterWebService would add an allow-all entry every time the service initialized. At this point MasterWebService still adds the allow-all entry if there are no authentication entries defined (e.g., `$VOLTTRON_HOME/auth.json` is empty).

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/674)


craig8:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/volttron/volttron/674)*

PR #684, opened 2016-08-05
Added demo and customized install instructions for this branch

craig8:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/684)


craig8:
Not gonna do it as we don't have the right naming for this.  It is going to be added to a diffferent branch.

PR #694, opened 2016-08-11
Improvements to bacnet config grabber

kmonson:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/694)


craig8:
Remind me why we left so much commented out code in this file?

kmonson:
It's code for something that I have no way of testing right now. I know that it's broken but I can't fix it so I just skip that potential path for now. When/if we get a device to validate/fix that code it will go right back in.

PR #700, opened 2016-08-16
Race condition fix

mikeroup:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/700)


mikeroup:
Ugh looks like it still might be happening.

craig8:
The tests pass now...since you closed this after you comment I would assume we are good?

mikeroup:
Had another install timeout. Still happens just less frequently.

PR #702, opened 2016-08-17
add exception to set auth event before router is started

mikeroup:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/702)


craig8:
Woot finally!  Thanks for your efforts @mikeroup

Issue #680, opened 2016-08-02
CAgent example doesn't work from command line


jhaack:
Setup needs to include c lib so the agent can run without manually copying files

PR #703, opened 2016-08-18
Fix c agent install #680

mikeroup:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/703)


examples/CAgent/c_agent/agent.py
```diff
@@ -69,7 +69,8 @@
 __version__ = 1.0
 
 '''This example agent calls functions from a shared object via
```
craig8: docstrings should be """
https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

examples/CAgent/c_agent/agent.py
```diff
@@ -69,7 +69,8 @@
 __version__ = 1.0
```
craig8: __version__ should be a string.
https://www.python.org/dev/peps/pep-0396/#specification

mikeroup:
The links on that page are currently broken.

Issue #710, opened 2016-08-26
Make AlerterAgent more generic


craig8:
Let us be able to configure this on the fly so we can call a reconfigure rpc function with topics that we want to alert on.

PR #718, opened 2016-08-29
Add rpc methods to alerter #710

mikeroup:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/718)

### DIFF
--- a/services/core/AlertAgent/IDENTITY
+++ b/services/core/AlertAgent/IDENTITY
@@ -1,0 +1,1 @@
+platform.alerter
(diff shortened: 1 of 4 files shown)


craig8:
When I mentioned reconfiguration this isn't what I had in mind, however this does make it more flexible.  I was just meaning have a function that would take return value of load_config(file) as a parameter.

So

````
    def configure(**kwargs):
        # watch all topics in here

I think that might be a good thing to add once the config store goes live, but otherwise I like what you have done here.

PR #721, opened 2016-08-29
Added volttron_home being passed to service agents

craig8:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/721)

### DIFF
--- a/volttron/platform/agent/utils.py
+++ b/volttron/platform/agent/utils.py
@@ -210,12 +210,15 @@ def vip_main(agent_class, identity=None, **kwargs):
         identity = os.environ.get('AGENT_VIP_IDENTITY', identity)
         address = os.environ.get('VOLTTRON_VIP_ADDR')
         agent_uuid = os.environ.get('AGENT_UUID')
+        volttron_home = os.path.abspath(get_home())
+
         if not address:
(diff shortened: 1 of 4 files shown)


craig8:
@zbeech I am adding you for comment on this.  I was going to add volttron_home to the AuthFile to get rid of the global dependence on VOLTTRON_HOME.  The goal is agents that start with the platform or installed on the platform will have volttron_home sent to them as an argument.  So the only one who should actually use the os.environ.get('VOLTTRON_HOME') is on the server and in vip_main.

So how do we modify your AuthFile function to make this happen?

zbeech:
@craig8 I think we should add RPC calls to AuthService that provide access to AuthFile functionality, rather than calling AuthFile directly. That way AuthService can provide VOLTTRON_HOME to AuthFile.

craig8:
@zbeech I agree...actually I am kinda surprised that we haven't done that already.  The thing is that blocking the caller with the privilege would require that the local interface be tcp based, which isn't a game killer just something that we need to be aware of.  Say instead of using ipc sockets we default to tcp://127.0.01:22916 or something. 

@kmonson  do you have any thought's on having a tcp address as "local" and doing away with ipc?

craig8:
Merged without pushing to craig8/vip-main

PR #733, opened 2016-09-03
Added code for restricting vip identities

craig8:
Added documentation for the valid characters for identities.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/733)

### DIFF
--- a/docs/source/specifications/agent-vip-id.rst
+++ b/docs/source/specifications/agent-vip-id.rst
@@ -1,7 +1,7 @@
 Agent VIP IDENTITY Assignment Specification
 =====================================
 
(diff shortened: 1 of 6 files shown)


craig8:
Well that's not good...seems I broke all the tests ;).  I will fix and then let @kmonson know when it's ready.

craig8:
Ok that's better go for it @kmonson

kmonson:
I don't think we should try to enforce the identity rules in Agent or vip_main. We can mention in the spec that it's not recommended and that some services may not work for the Agent and leave it at that. We could put in a warning if you want. 

We should leave in all the other sanity checks at installation time of course.

craig8:
I though that they s was what we discussed last week.  Did our thinking change?

kmonson:
I don’t think we ever discussed where we were going to enforce this. Let’s talk about it tomorrow at work.

craig8:
Closing this based upon feedback from @kmonson.  I will modify and re-submit.

PR #736, opened 2016-09-08
Feature/config store

kmonson:
Agent side of the config store.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/736)


volttron/platform/store.py
```diff
@@ -284,9 +233,40 @@ def set_config(self, config_name, contents, trigger_callback=False):
...
+
+        #We need to create store and lock if it doesn't exist in case someone tries to add
+        # a configuration while we are sending the initial state.
+        agent_store = self.store.get(identity)
+
+
+        if agent_store is None:
+            # Initialize a new store.
+            store_path = os.path.join(self.store_path, identity + store_ext)
+            store = PersistentDict(filename=store_path, flag='c', format='json')
```
craig8: What will happen if someone puts an invalid path say for instanc e a <, > character will your store path blow up?

volttron/platform/vip/agent/subsystems/configstore.py
```diff
@@ -0,0 +1,375 @@
...
+from .base import SubsystemBase
+from volttron.platform.storeutils import list_unique_links, check_for_config_link
+
+from collections import defaultdict
+from copy import deepcopy
+
+"""The heartbeat subsystem adds an optional periodic publish to all agents.
+Heartbeats can be started with agents and toggled on and off at runtime.
+"""
+
```
craig8: I don't understand why this comment is here.

Can you add some documentation to this file.  I know most of the methods are documented via the spec, however that doesn't get added to our api documentation output.

One more thing of note, should we have "public" methods first in the file or second...I know it really doesn't make a difference, but thought I would ask your opinion

craig8:
@kmonson can you address these please.  I will get it downloaded and try to implement one of the stores and then go ahead and merge it in.

volttron/platform/vip/agent/subsystems/configstore.py
kmonson: Bad copy/paste... 

I like to put my helper functions at the top as I usually write them first.

I can add some doc strings.

volttron/platform/store.py
kmonson: Yeah, right now it will probably blowup. Should only affect the bad agent and not cause any lasting harm.

PR #741, opened 2016-09-13
Missed raising exception during install if idenitty is invalid

craig8:
This should reference #734

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/741)


craig8:
OUCH this will break everything so we aren't going to do this on install just now!

Issue #251, opened 2015-12-04
ActuatorAgent requests should be based on a platform provided ID


jhaack:
Actuator Agent uses a self reported id from agents. It should instead be based on a platform provided ID to prevent agents from spoofing a reservation.

jhaack:
- Make pub/sub optional for ActuatorAgent set/get point, revert, etc. Continue publishing general information (schedules)
- Make RPC use VIP identity for purposes of tracking reservations. RequestorID is deprecated but can be included in headers as informational.

PR #744, opened 2016-09-13
Use VIP ids for actuator schedules #251

mikeroup:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/744)


docs/source/core_services/service_agents/actuator/ActuatorScheduleResponse.rst
```diff
@@ -10,7 +10,7 @@ ActuatorAgent will respond on the topic
     #python
     {
         'type': <'NEW_SCHEDULE', 'CANCEL_SCHEDULE'>
-        'requesterID': <Agent ID from the request>,
+        'requesterID': <Agent VIP identity from the request>,
```
craig8: Is the requesterID required?  Perhaps mark as legacy or something and not include. @jhaack thoughts on this?

services/core/ActuatorAgent/actuator/agent.py
```diff
@@ -426,7 +427,7 @@
 
     {
         'type': 'CANCEL_SCHEDULE',
-        'requesterID': <Agent ID associated with the preempted Task>,
+        'requesterID': <VIP id associated with the preempted Task>,
```
craig8: Should probably spell the identity out like you do in most other places

services/core/ActuatorAgent/tests/test_actuator_pubsub.py
```diff
@@ -560,6 +560,7 @@ def test_schedule_empty_task(publish_agent, cancel_schedules):
 
 
 @pytest.mark.actuator_pubsub
+@pytest.mark.xfail(reason="Request ids are now ignored.")
```
craig8: I am not sure I like how many tests are xfailed within this module now.

docs/source/core_services/service_agents/actuator/ActuatorScheduleResponse.rst
kmonson: If we remove the argument then every module that works with the AA will break.

Eventually we'll add a replacement RPC function (not now as the interface of the RPC function needs to change in other ways) and we'll mark this method deprecated.

services/core/ActuatorAgent/tests/test_actuator_pubsub.py
kmonson: Should probably remove tests that verify the identity instead of marking them xfail.

docs/source/core_services/service_agents/actuator/ActuatorScheduleResponse.rst
craig8: Ahh good lets not remove that then :)

PR #749, opened 2016-09-14
Fixed code path copy paste error.

craig8:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/749)


scripts/bacnet/proxy_grab_bacnet_config.py
```diff
@@ -151,13 +149,13 @@ def process_object(address, obj_type, index, max_range_report, config_writer):
     subondinate_list_property = get_datatype(obj_type, 'subordinateList')
```
kmonson: Comment out 149 to 159 and the follow on code will do the right thing.

craig8: Done

PR #751, opened 2016-09-15
Add example systemd unit file

peter-nordquist:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/751)


scripts/admin/volttron.service
```diff
@@ -0,0 +1,16 @@
+# VOLTTRON SystemD unit
```
craig8: Which operating systems use SystemD.  I know ubuntu based systems do, but do others?

peter-nordquist: Ubuntu 16.04 and Centos 7 both are currently using System D

PR #755, opened 2016-09-19
Feature/auth ipc

zbeech:
This fork enables encryption and authentication over IPC.  Default encryption-related behavior was changed in in the platform and Agent Core to make this work:

### Developer mode
- Encryption is enabled by default for all TCP and IPC sockets
- To disable encryption, pass `developer_mode=True` to the `Agent` class
- The `--developer_mode` option was already part of `volttron`, and it is now available to `volttron-ctl` 
- Agents that are installed when `volttron-ctl` is in developer mode will also be in developer mode.

### Automatic key retrieval 
If encryption keys are not passed to the `Agent` class, then it will automatically fetch these keys if it can. It will get the public and secret keys from `KeyStore` and the server key from `KnownHosts`.

### Automatic key authorization

- When the platform starts it automatically authorizes its own public key. Local `volttron-ctl` instances use the platform's key (by default) to authenticate to the platform. This saves us the trouble of creating and managing a separate `volttron-ctl` key. Because a local `volttron-ctl` instance can read and change the platforms key anyway, it might as well use that key.
- When the platform starts it automatically adds its own public key to the `known_hosts` file. This provides an easy way for installed agents to know what server key to use when connecting to the platform.
- When an agent is installed its public key is automatically generated and added to the auth file.
- When an agent is removed its public key is automatically removed from the auth file.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/755)


craig8:
the known_host file is to be what read by agents connected to the platform?  Or is there an api somewhere?

Also, the functionality exists for the agents that are connected to the platform to get at that information through the querying of the router.  Though this might not be the same thing.

zbeech:
@craig8 In Agent Core (https://github.com/zbeech/volttron/blob/feature/auth_ipc/volttron/platform/vip/agent/core.py#L503) the known_hosts file is read via the `KnowHostStore` class.

In this case, the known_hosts file is a way for agents to get the platform's public key before they connect to the platform (because they can't connect to the platform without knowing its public key).

volttron/platform/auth.py
```diff
@@ -634,6 +634,18 @@ def add(self, auth_entry, overwrite=True):
             entries.append(auth_entry)
             self._write(entries, groups, roles)
 
+    def remove_by_credentials(self, credentials):
+        """
```
craig8: function comments should start right after the first """

volttron/platform/keystore.py
```diff
@@ -156,6 +156,11 @@ def isvalid(self):
 class KnownHostsStore(BaseJSONStore):
     '''Handle storage and retrival of known hosts'''
```
craig8: Double quotes should be used in class and function documentation.

volttron/platform/vip/agent/core.py
```diff
@@ -429,28 +429,40 @@ def __init__(self, owner, address=None, identity=None, context=None,
...
-        if publickey is None or secretkey is None:
-            publickey, secretkey = self._get_keys()
-        if publickey and secretkey and serverkey:
-            self._add_keys_to_addr(publickey, secretkey, serverkey)
+        self._set_keys()
 
-        if publickey is None:
-            _log.debug('publickey is None')
-        if secretkey is None:
-            _log.debug('secretkey is None')
```
craig8: should probabably add a line so that it's easy to figure out that you are within the core class.

zbeech:
@craig8 I need to add an option to the build_agent test fixture to force it to use IPC. After I get that done, all the authorization tests are here: https://github.com/zbeech/volttron/blob/feature/auth_ipc/volttrontesting/platform/auth_test.py

zbeech:
Review status: 0 of 15 files reviewed at latest revision, 2 unresolved discussions.

---

*[volttron/platform/auth.py, line 638 at r1](https://reviewable.io:443/reviews/volttron/volttron/755#-KS37UPcLKOoSyp1tc8u-r1-638:-KS7SeT_w-HwzlTUnsgY:b-896fix) ([raw file](https://github.com/volttron/volttron/blob/530b371432f69c074a2fac4b011dc52fcc619454/volttron/platform/auth.py#L638)):*
<details><summary><i>Previously, craig8 (Craig) wrote…</i></summary>
> function comments should start right after the first """
</details>
Done.

---

*[volttron/platform/keystore.py, line 157 at r1](https://reviewable.io:443/reviews/volttron/volttron/755#-KS37UPcLKOoSyp1tc8w-r1-157:-KS7SgsHXkfvdk-boMQV:b-896fix) ([raw file](https://github.com/volttron/volttron/blob/530b371432f69c074a2fac4b011dc52fcc619454/volttron/platform/keystore.py#L157)):*
<details><summary><i>Previously, craig8 (Craig) wrote…</i></summary>
> Double quotes should be used in class and function documentation.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/volttron/volttron/755)*

PR #756, opened 2016-09-20
let installs skip the channel subsystem for local installs

mikeroup:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/756)


craig8:
Thanks @mikeroup

craig8:
So the question becomes how does one get the file:// to the control?

craig8:
please add file://$WHEEL to the pack_install script so that we utilize this method

PR #768, opened 2016-09-26
Disable watch_file() if inotify is not supported on platform.

ffreckle:
inotify is not supported on OS X/MacOS. This change
catches the library import error and changes watch_file()
to log warning that runtime changes are not supported on
the platform.  In the future, OS X FS Events API could be
used to implement the same functionality on OS X/MacOS.

This has been tested on Ubuntu 14.04 and OS X 10.10.5.
Let me know if you would prefer to handle this platform
dependency in some other way.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/768)

### DIFF
--- a/volttron/platform/agent/utils.py
+++ b/volttron/platform/agent/utils.py
@@ -77,7 +77,14 @@ from dateutil.parser import parse
 from dateutil.tz import tzutc
 from tzlocal import get_localzone
 from zmq.utils import jsonapi
-from ..lib.inotify.green import inotify, IN_MODIFY
+
+try:
+    from ..lib.inotify.green import inotify, IN_MODIFY
+except AttributeError:
+    # inotify library is not available on OS X/MacOS.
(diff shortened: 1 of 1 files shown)


craig8:
Thank you @ffreckle!

zbeech:
We might eventually be able to remove inotify all together. It looks like it is it is only used for authorization-related files (auth.json, protected_topics.json). If we provide sufficient APIs for changing these files (i.e., via volttron-ctl and RPC calls) then we could alert the appropriate services (`AuthService` and `PubServices`) that these files have changed when the API is used. Same applies if we end up using the `ConfigStoreService` in place of these JSON files.

kmonson:
I don't see the harm in accepting this right after we release 4.0 just to get OSX based developers up and running. @zbeech is right in that watch_file will be obsolete once auth services start using the configuration store which will probably happen Q1 or Q2 of FY17.

PR #778, opened 2016-09-28
Updated to 15.4 for pyzmq

craig8:
Refroze requirements file.

In addition, I updated the version number of the package volttron.platform.  And made that the default in the setup.py file

I also added it as the output from show-config.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/778)


jhaack:
Tests pass with modified run-test.sh script

PR #780, opened 2016-09-28
Added get_home and get_address to the grabbers so they will actually …

craig8:
…work.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/780)


kmonson:
Good enough.

Issue #792, opened 2016-09-30
VOLTTRON process from volttron-cfg not closing on exit


jhaack:
I have observed the volttron process from volttron-cfg still running after exiting the wizard.

This then causes errors trying to start up the platform to see your results which could be very confusing for users.

jhaack:
Have not observed since latest updates.

PR #794, opened 2016-09-30
volttron-cfg fix/speedup

mikeroup:
The volttron-ctl commands were timing out. Since the ipc auth additions if volttron is started with developer mode volttron-ctl needs to use it too.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/794)


mikeroup:
Probably also fixes #792

craig8:
@jhaack can you verify if this fixes #792 
Thanks

PR #806, opened 2016-10-04
File watch publisher

poorva1209:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/806)


volttron/platform/agent/utils.py
```diff
@@ -461,6 +461,21 @@ def watch_file(fullpath, callback):
...
+    else:
+        with inotify() as inot:
+            inot.add_watch(dirname, IN_MODIFY)
+            for event in inot:
+                if event.name == filename and event.mask & IN_MODIFY:
+                    callback(fullpath)
+
 
 def create_file_if_missing(path, permission=0o660, contents=None):
     dirname = os.path.dirname(path)
```
craig8: Do we use this? anywhere?

poorva1209: yeah, we are using it in FileWatchPublisher/filewatchpublisher/agent.py :
```
@Core.receiver('onstart')
    def starting(self, sender, **kwargs):
        _log.info("Starting FileWatchPublisher agent")
        for item in self.config:
            file = item["file"]
            self.core.spawn(watch_file_with_fullpath, file, self.read_file)
```

craig8: Nevermind I was actually talking about the create_file_if_missing function :) My bad not your code

craig8:
Can you add documentation on how to use this to the docs/service_agents? http://volttron.readthedocs.io/en/develop/core_services/service_agents/index.html

poorva1209:
Sure, will do it right away.

poorva1209:
Done....agent added to service_agent list.

services/core/FileWatchPublisher/IDENTITY
```diff
@@ -0,0 +1 @@
+platform.filewatchpublisher
```
craig8: add {n} to the end of this...That will allow us to install more than one of these agents...

services/core/FileWatchPublisher/filewatchpublisher/agent.py
```diff
@@ -0,0 +1,159 @@
...
+    Example configuration:
+
+    .. code-block:: python
+
+        {
+	        "publish_file": [
+                {
+                    "file": "/var/log/syslog",
+                    "topic": "platforms/syslog",
+                },
```
craig8: use platform/syslog as the the topic for these examples
I committed some of those above, but didn't in this one.

services/core/FileWatchPublisher/filewatchpublisher/agent.py
```diff
@@ -0,0 +1,159 @@
...
+                {
+                    "file": "volttron.log",
+                    "topic": "platforms/log",
```
craig8: I don't think we want to include the volttron log to be published....seems like that wouldn't be productive.  so lets not do that here.

services/core/FileWatchPublisher/filewatchpublisher/agent.py
```diff
@@ -0,0 +1,159 @@
...
+        self.file_end_position = {}
+        for item in self.config:
+            file =  item["file"]
+            self.file_topic[file] = item["topic"]
+            with open(file, 'r') as f:
+                self.file_end_position[file] = self.get_end_position(f)
+
+    @Core.receiver('onstart')
+    def starting(self, sender, **kwargs):
+        _log.info("Starting FileWatchPublisher agent")
```
craig8: instead of embedding the class name you can use self.__class__.__name__ or self.__name__.  That way if we change it later we won't have to change it here.  Note the self.__name__ is only relevant if you set the instance on line 84 to the __name__ of the class.

services/core/FileWatchPublisher/IDENTITY
poorva1209: Added.

services/core/FileWatchPublisher/filewatchpublisher/agent.py
poorva1209: Done.

services/core/FileWatchPublisher/filewatchpublisher/agent.py
poorva1209: Changed it to _log.info("Starting "+self.__class__.__name__+" agent")

docs/source/core_services/service_agents/file_watch_publisher/FileWatchPublisher-Agent.rst
```diff
@@ -0,0 +1,30 @@
...
+Files and topics should be provided in the configuration.
+
+Configuration
+-------------
+
+A simple configuration for FileWatchPublisher with two files to monitor is as follows:
+
+::
+
+    [
```
craig8: Should the config file be an object itself with the array as a property value?  I know it's valid json but it seems strange not to have the outer { in configurations.

craig8:
We need to handle the case that the file doesn't currently exist.  We need to either create the file or have a better output than the stack trace that is in the log.   I think rather than creating the file, capture the IOError and making a more appropriate log message would be good.

PR #808, opened 2016-10-04
Feature/config store

kmonson:
Config store docs and updates to how config store handles lack of support on the platform.

Also turned off config store support on forwarder child agent.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/808)


kmonson:
Bleh, github...

Issue #810, opened 2016-10-05
Fix sqlite historian configuration files


jhaack:
platform config is missing platform.historian identity
regular historian has DECTYPES line which breaks historian.

Issue #811, opened 2016-10-05
Identity not being picked up from configuration file


jhaack:
"identity" is not being used in old style configuration files (such as platform.historian). 

This will cause confusion with deployments not done through volttron-cfg.

Either pick identity up out of old configuration files or highlight this breaking change.

kmonson:
We will highlight the breaking change. How do we want to do that?

craig8:
I think that this is a change that we can do, however I would like to be able to modify the historians so that they can read that first and then pass them to the config store.  This ability should be deprecated and will be removed at the 4.1 release.  That gives people time to fix this before we go on.

craig8:
We should document this as a future pain point in the release notes.  and in readthedocs as well.

kmonson:
There is a TERRIBLE AND AWFUL hack where we pop the identity given to us by vip_main and supply our own. 

The reason why I hate this is that the VIP IDENTITY given by 

    volttron-ctl status

is no longer ground truth for that agent. As the historian does not currently support the configuration store this is not as terrible as it could be.  If we do put in this hack it must be removed when configuration support is added historians.

jhaack:
2016-10-05 21:43:22,044 (sqlhistorianagent-3.5.0 8084) volttron.platform.vip.agent.core INFO: Connected to platform: router: 92118517-0c94-448a-a52c-655c4cd15966 version: 1.0 identity: platform.historian
2016-10-05 21:43:22,044 (sqlhistorianagent-3.5.0 8084) sqlhistorian.historian INFO: Starting historian with identity: platform.historian
2016-10-05 21:43:22,074 (sqlhistorianagent-3.5.0 8084) <stderr> ERROR:     self.core.identity).get(timeout=2)

craig8:
That means you have one installed already.  I believe

jhaack:
Looks like the issue is that the changes in this pull fix it in the agent, looks like it is being started with the correct identity. However, the IDENTITY file is set in packaging which does not pick up that change.

PR #812, opened 2016-10-05
#810 Fix and update sqlite config files

jhaack:
#810 Add platform.historian identity back into sqlite platform historian config. Take out bad entry in regular sqlite config.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/812)


craig8:
This doesn't address #811 though correct?

jhaack:
Correct, this is simply to to fix the config files. #811 will then use the fixed config

PR #814, opened 2016-10-05
Added 'auth publickey' command to volttron-ctl

zbeech:
This command uses the "filterable" style similar to the `list` and `status` commands.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/814)


volttron/platform/control.py
```diff
@@ -934,6 +934,47 @@ def update_auth(opts):
         _stderr.write('ERROR: %s\n' % err.message)
 
 
+def _show_filtered_agents(opts, field_name, field_callback, agents=None):
```
craig8: Please document what inputs and outputs of this function are...and why this is necessary.

zbeech: I documented that function in a following commit. I put that logic in its own function because I noticed that same pattern is used a couple times (e.g., `list_agents`, `status_agents`). I didn't want to change `list_agents` and `status_agents` in this pull request, but I would like to later.

craig8:
I was actually expecting to have the output of this be like the volttron-ctl auth serverkey, however I like this.

PR #824, opened 2016-10-06
Performance improvements for mongo historian query

schandrika:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/824)


volttron/platform/agent/base_historian.py
```diff
@@ -304,9 +304,9 @@ def _create_subscriptions(self):
         subscriptions = [
             (topics.DRIVER_TOPIC_BASE, self._capture_device_data),
             (topics.LOGGER_BASE, self._capture_log_data),
-            (topics.ACTUATOR, self._capture_actuator_data),
+            #(topics.ACTUATOR, self._capture_actuator_data),
```
craig8: Why are we commenting this out?

services/core/MongodbHistorian/tests/test_mongohistorian.py
```diff
@@ -612,16 +646,17 @@ def test_empty_result(volttron_instance, database_client):
...
 
         lister = volttron_instance.build_agent()
 
         result = lister.vip.rpc.call(
             'platform.historian',
             'query',
-            topic=BASE_ANALYSIS_TOPIC[9:]+'/FluffyWidgets').get(timeout=5)
-        print ("query result:" ,result)
+            topic=BASE_ANALYSIS_TOPIC[9:] + '/FluffyWidgets').get(timeout=5)
+        print ("query result:", result)
```
craig8: remove print messages

volttron/platform/agent/base_historian.py
kmonson: Yeah, we're just going to punt for now. We need some sort of basehistorian spec. I was kind of winging it when I wrote it the first time.

PR #849, opened 2016-10-11
Added 'volttron-ctl upgrade' command

zbeech:
* `volttron-ctl upgrade` removes and installs agent while preserving the agent's existing authorization entry in `$VOLTTRON_HOME/auth.json`
 * Changed `KeyStore` methods `public` and `secret` to properties (this helped with preserving authorization entry and it is more pythonic)
 * Removed unused imports of `KeyStore`, `AuthEntry`, and `AuthFile`

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/849)

### DIFF
--- a/services/core/FailoverAgent/tests/test_failover.py
+++ b/services/core/FailoverAgent/tests/test_failover.py
@@ -44,8 +44,8 @@ def tcp_to(instance):
     return "{}?serverkey={}&publickey={}&secretkey={}".format(
         instance.vip_address,
         instance.serverkey,
-        key.public(),
-        key.secret())
+        key.public,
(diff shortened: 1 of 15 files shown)


craig8:
I envision that you shouldn't change the method to properties as that change will break lots of things unless you did it all over the place.

Also, that is one more backward compatibility.  Note I do like that it is a property.  Just make sure that you have all of those changed.

craig8:
Of course I say this without looking at your code which has ks.public instead of ks.public() all over the place.:)

craig8:
Thanks @zbeech

PR #858, opened 2016-10-11
Feature/config store test

kmonson:
Configuration store tests. 

Also sorts list results from some manage calls.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/858)

### DIFF
--- a/volttron/platform/store.py
+++ b/volttron/platform/store.py
@@ -214,13 +214,17 @@ class ConfigStoreService(Agent):
 
     @RPC.export
     def manage_list_configs(self, identity):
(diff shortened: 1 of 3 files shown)


volttron/platform/store.py
```diff
@@ -214,13 +214,17 @@ def manage_delete_store(self, identity):
 
     @RPC.export
     def manage_list_configs(self, identity):
-        return self.store.get(identity, {}).get("store", {}).keys()
+        result = self.store.get(identity, {}).get("store", {}).keys()
+        result.sort()
```
craig8: Is sorting necessary or jus making it easier to test?

Issue #765, opened 2016-09-23
Greenlets not getting cleaned up


mikeroup:
I'm mostly interested in ones that have dictionaries like the following. These can be easily created by `volttron-ctl status`. There are greelets that look like this before control is ever called which makes me think it may be an agent/core issue and not a control issue.

```
{
    '_notifier': <callback at 0x7fba7038e990 stopped>,
    '_start_event': <gevent.greenlet._dummy_event object at 0x7fba75ad9f50>,
    'value': None,
    '_links': deque([]),
    '_exc_info': (None, None, None)
}
```

I have a service like the one below to look at greenlets over time. The number of greenlets only increases. I used print instead of the logging module so that I can run `$ volttron` and have only this output.
``` python
import gc
from greenlet import greenlet
import pdb

class MemoryAnalysis(Agent):
    def __init__(self, *args, **kwargs):
        super(MemoryAnalysis, self).__init__(*args, **kwargs)
        self.iteration = 0
        self.oldstate = set(self.get_greenlets())

    @Core.periodic(15)
    def show_usage(self):
        print("================================================================================")
        print("ITERATION {}".format(self.iteration))
        self.iteration += 1

        # clean up unneeded objects
        gc.collect()

        # get all greenlets
        newstate = set(self.get_greenlets())
        print("THERE ARE {} GREENLETS".format(len(newstate)))

        # put previously untracked greenlets in diff
        diff = newstate - self.oldstate
        self.oldstate = newstate

        # pdb.set_trace()

    def get_greenlets(self):
        return [g for g in gc.get_objects() if isinstance(g, greenlet)]
```

mikeroup:
After some more investigating I generated the following graph with the objgraph module. Most of it seems to be noise from the necessary local references and pdb. I highlighted an interesting part.

![outline](https://cloud.githubusercontent.com/assets/4081515/18819415/7b846c38-8345-11e6-8028-1118b02eae8a.png)

mikeroup:
I used the inspect module to make greenlets record the stack when they are created. Looks like leftover greenlets are going through this execution pathway. 

```
(<frame object at 0x7f7bac10bbf0>, '/home/mike/volttron/env/local/lib/python2.7/site-packages/gevent/greenlet.py', 396, 'spawn_later', ['        g = cls(*args, **kwargs)\n'], 0)
[...]

PR #869, opened 2016-10-14
remove spawn later link #765

mikeroup:
This is a pretty common pattern:
```
greenlet = gevent.spawn_later(seconds, func, *args, **kwargs)
self.greenlet.link(lambda glt: greenlet.kill())
```
we could probably remove it in more places but this one is the most noticeable.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/869)

### DIFF
--- a/volttron/platform/vip/agent/core.py
+++ b/volttron/platform/vip/agent/core.py
@@ -343,6 +343,12 @@ class BasicCore(object):
         self.greenlet.link(lambda glt: greenlet.kill())
         return greenlet
 
+    def spawn_later(self, seconds, func, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


kmonson:
Is it the link that causes it to stick around? I thought that was to make sure a greenlet was killed if the main greenlet was stopped.

mikeroup:
The extra reference keeps them from getting garbage collected after they finish.

kmonson:
We will need to setup some mechanism that unlinks the callback after the greenlet has finished. Using weakref will not be sufficient. It will allow the greenlet to be garbage collected, but the linked callback will still be kept around indefinitely.

PR #873, opened 2016-10-14
Documentation updates for historian and aggregate historian

schandrika:
Documentation updates for historian and aggregate historian. Also fixed a bunch of fixes so that sphinx doesn't show any ERROR and SEVERE level errors when "make html" is run

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/873)


craig8:
Can you remove references to datetime.utcnow().isoformat() + 'Z' and use our custom function for this instead?

docs/source/core_services/historians/Historian-Topic-Syntax.rst
```diff
@@ -55,26 +58,51 @@ https://github.com/VOLTTRON/volttron/blob/develop/examples/ExampleSubscriber/sub
...
         [52.5, {'units': 'F', 'tz': 'UTC', 'type': 'float'}]
     }
 
     # ALL TOPIC
-    # Messages contains a two element list.  The first element contains a dictionary of all points 
-    # under a #specific parent.  While the second element contains a dictionary of meta data for each of the specified points.  For example devices/pnnl/building/OutsideAirTemperature and 
-    # devices/pnnl/building/MixedAirTemperature ALL message would be created like:
+    # Messages contains a two element list.  The first element contains a
+    # dictionary of all points under a #specific parent.  While the second
+    # element contains a dictionary of meta data for each of the specified
```
craig8: the # should be removed in front of specifc

volttron/platform/dbutils/mysqlfuncts.py
```diff
@@ -213,13 +219,7 @@ def setup_aggregate_historian_tables(self, meta_table_name):
     def query(self, topic_ids, id_name_map, start=None, end=None, skip=0,
               agg_type=None,
               agg_period=None, count=None, order="FIRST_TO_LAST"):
-        """This function should return the results of a query in the form:
```
craig8: Should we include pointers to the base in each of the functtions?

schandrika: I tried to do it with an event function on sphinx but couldn't. I have add this manually. That's why added it at top level. It doesn't look bad in the resulting html since the only documentation child classes have are class level and main so its quite prominent.

docs/source/core_services/historians/Historian-Topic-Syntax.rst
schandrika: Done

volttron/platform/dbutils/mysqlfuncts.py
craig8: ok sounds good.

docs/source/applications.rst
```diff
@@ -1,9 +0,0 @@
-Applications
```
craig8: This actually should probably link to a read the docs for volttron-applications....instead of deleting...that doesn't exist though so this works :)

craig8:
Thanks @schandrika

PR #875, opened 2016-10-14
More documentation changes

schandrika:
Aggregate historian and historian implementations will now show docstrings from base class in readthedocs if not docstring exists in child class methods. 
More formatting changes in docstring

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/875)


services/core/MongodbHistorian/tests/test_prod_query_mongo.py
```diff
@@ -0,0 +1,367 @@
+import pytest
+from  datetime import datetime
+#building name replaced. replace before testing
```
craig8: So the topic would be BUILDNG1 not Building1.  But since we are case insensitive this works as written?

volttron/platform/agent/base_aggregate_historian.py
```diff
@@ -290,8 +284,8 @@ def _init_agg_group(self, agg_group, agg_time_period):
             _log.debug("End of loop in init_agg_group. ids {}".format(
                 data['topic_ids']))
 
-    def _collect_aggregate_data(self, collection_time, agg_time_period,
-                                use_calendar_periods, points):
+    def collect_aggregate_data(self, collection_time, agg_time_period,
```
craig8: Why are we changing the _ of these.  Are these meant to be called from outside this module?

schandrika: because sphinx won't pull docstring from _ methods

services/core/MongodbHistorian/tests/test_prod_query_mongo.py
craig8: BUILDING1 is the anonymized version in our code.  This isn't a big issue.

Issue #880, opened 2016-10-17
missing code


nikithark:
[http://volttron.readthedocs.io/en/develop/devguides/agent_development/TestAgent.html](url)

import code for setup.py missing -  from setuptools import setup, find_packages

PR #885, opened 2016-10-17
add import to testagent setup, fix logging method #880

mikeroup:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/885)


craig8:
Is this for an issue...if so please mention it somewhere so I can track it.  If not why not?

mikeroup:
For issue #880

PR #901, opened 2016-10-18
update connection to make sure we close down

craig8:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/901)


craig8:
@mikeroup test please

Issue #896, opened 2016-10-17
Improve base page of documentation


jhaack:
Landing page for readthedocs needs to have quick jumping off points to help people get started.

PR #903, opened 2016-10-18
Updated main docs page to have more information

jhaack:
#896 More informative base page. Also moved deprecated examples.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/903)


docs/source/README.rst
```diff
@@ -1,27 +1,50 @@
...
+    volttron -vv -l volttron.log&
+
+This enters the virtual Python environment and then starts the platform in debug (vv) mode with a log file named
+volttron.log.
+
+Next, start an example listener to see it publish and subscribe to the message bus:
+
+    scripts/core/make-listener
+
+This script handles several different commands for installing and starting an agent after removing an old copy. This
```
craig8: I don't think this rendering is what you had in mind.
index-error-rendering

docs/source/README.rst
```diff
@@ -33,9 +56,10 @@ Support
 -------
 
 If you are having issues, please let us know.
```
craig8: STACKOVERFLOW!!!!!

docs/source/specifications/configuration-store.rst
```diff
@@ -1,3 +1,5 @@
+.. _ConfigurationStore:
+
 Agent Configuration Store
```
craig8: Seems like hte other pages have a
============================================
Title
============================================

Perhaps we should keep that constistancy or change to the other way?

craig8: hmm you need to look at the edit of that to see the difference ha

jhaack: Going to have an editor go through and make pages consistent.

PR #905, opened 2016-10-19
Update ListenerAgent.rst

nikithark:
A little unsure about these edits. Corrected methods to reflect the new version. @Core.periodic decorator does not exist.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/905)

### DIFF
--- a/docs/source/supporting/examples/ListenerAgent.rst
+++ b/docs/source/supporting/examples/ListenerAgent.rst
@@ -23,7 +23,7 @@ definition.
 
     from volttron.platform.vip.agent import Agent, Core, PubSub, compat
     from volttron.platform.agent import utils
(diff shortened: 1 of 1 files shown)


kmonson:
These lines are no longer true and should be removed.

    ListenerAgent publishes a heartbeat message so it will use the PublishMixin. It also extends
    BaseAgent to get the default functionality. When creating agents, Mixins should be first in the class definition.

The stuff about setting up logging (starting with "utils.setup_logging()") is formatted incorrectly. All lines starting with a pipe need to be fixed and the pipe removed. There are several more cases in the file. 

The line

    message = compat.unpack_legacy_message(headers, message)

needs to be indented 4 more spaces.

Issue #922, opened 2016-10-19
Fix version numbers for core agents with significant changes.


kmonson:
We need to make sure that all core agents with significant changes since 3.5 have had their version number bumped.

PR #924, opened 2016-10-20
add version to failover agent.py

mikeroup:
for #922

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/924)

### DIFF
--- a/services/core/FailoverAgent/failover/agent.py
+++ b/services/core/FailoverAgent/failover/agent.py
@@ -67,6 +67,7 @@ from volttron.platform.vip.agent.connection import Connection
 
 utils.setup_logging()
 _log = logging.getLogger(__name__)
+__version__ = '0.1'
 
 
 class FailoverAgent(Agent):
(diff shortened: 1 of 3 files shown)


services/core/FailoverAgent/failover/agent.py
```diff
@@ -67,6 +67,7 @@
 
 utils.setup_logging()
 _log = logging.getLogger(__name__)
+__version__ = '0.1'
```
craig8: Probably should be higher since this was out for a bit...Even though I doubt people have used it.

mikeroup: This is what Kyle said to do. Did you have a specific number in mind?

craig8: Kyle said that if would have thunk about it he would have had you bump it.  You can choose any number you like above 0.1.  I would however suggest 0.2.   But that's just me.  My guide would be http://semver.org/.

I might even be able to argue 0.1.1 as we changed the configuration interface.

mikeroup: I'll bump it to 0.2 when I add the alert publish.

PR #927, opened 2016-10-20
dds example without library included

mikeroup:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/927)


jhaack:
Could you put together a page talking about how to use this agent (including adding the dds library)?

examples/DDSAgent/README
```diff
@@ -0,0 +1,16 @@
+This agent example is meant to be run from this directory with:
+
```
craig8: From the command line, as opposed to installin

PR #998, opened 2016-10-27
print message instead of stack trace on zmq rebind 

mikeroup:
# 274

---

This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/998)


craig8:
What is 274? is that the number of the issue?

Issue #776, opened 2016-09-27
Update failing tests


jhaack:
Tests failing likely because they need to be updated
- [x] volttrontesting/testutils/test_platformwrapper.py::test_resinstall_agent[encrypted] FAILED
- [x] volttrontesting/testutils/test_platformwrapper.py::test_resinstall_agent[unencrypted] FAILED
- [x] volttrontesting/platform/packaging-tests.py::TestPackaging::test_can_create_package FAILED
- [x] volttrontesting/platform/packaging-tests.py::TestPackaging::test_can_extract_package FAILED
- [x] volttrontesting/platform/packaging-tests.py::TestPackaging::test_raises_error_if_agent_dir_not_exists FAILED

schandrika:
Unable to reproduce first 2. For the remaining 3 created new test module. Did not delete the older one as I want to keep it around as reference whenever I update the new one for restricted code related tests

PR #1000, opened 2016-10-27
Fix for #776

schandrika:
Fix for #776 Unable to reproduce test_platformwraper xfails so removed it. Created new test case for packaging.py

---

This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1000)


craig8:
Thank you @schandrika

PR #1007, opened 2016-10-31
do not add allow-all entry in MasterWebService, fixes #673

### LINKED ISSUE #673: MasterWebService adds 'allow-all' authentication entry

zbeech:
The removed code was not being run anyways, because the AuthFile will always have at least one entry (the platform itself).

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1007)

### DIFF
--- a/volttron/platform/web.py
+++ b/volttron/platform/web.py
@@ -189,18 +189,6 @@ class MasterWebService(Agent):
         super(MasterWebService, self).__init__(identity, address, **kwargs)
 
         self.bind_web_address = bind_web_address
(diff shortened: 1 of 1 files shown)


craig8:
Are we saying if the user does it, its fine but we aren't going to do it for them.  If so that works for me.

zbeech:
@craig8 Yes, this PR is only addressing MasterWebService automatically adding the allow-all entry.

craig8:
@zbeech Thankyou.

Issue #1014, opened 2016-11-01
move/remove uuid from the heartbeat


craig8:
Now that we have identities that are unique per platform.  We can get rid of the uuid from the heartbeat topic. 

In a vc context the vcp will modify the topic from heartbeat/identity to heartbeat/viphash/identity instead.

PR #1015, opened 2016-11-01
remove uuid from heartbeat publishes

mikeroup:
#1014

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1015)


craig8:
This doesn't include the identity anywhere.  We should have heartbeat/identity or heartbeat/name/identity but not heartbeat/name.  As that isn't unique on a given platform.

Issue #1020, opened 2016-11-03
Revamp ForwardHistorian


craig8:
The ForwardHistorian should only require the vip-address and serverkey in its configuration.  The agent when connecting should use the core.secretkey and core.publickey when connecting to the external platform.

PR #1029, opened 2016-11-08
Forwarder revamp

mikeroup:
For some reason this only seems to work if the know host is set #1020

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1029)

### DIFF
--- a/services/core/ForwardHistorian/config
+++ b/services/core/ForwardHistorian/config
@@ -2,6 +2,7 @@
 
     "agentid": "forwarder",
     "destination-vip": "ipc://@/home/volttron/.volttron/run/vip.socket",
+    "destination-serverkey": "",
     "required_target_agents": [],
     "custom_topic_list": [],   
     "services_topic_list": [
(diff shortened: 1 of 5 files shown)


services/core/ForwardHistorian/config
```diff
@@ -2,6 +2,7 @@
 
     "agentid": "forwarder",
     "destination-vip": "ipc://@/home/volttron/.volttron/run/vip.socket",
+    "destination-serverkey": "",
```
zbeech: This should be null rather than "" in configuration file. Otherwise it will be passed as an empty string which is an invalid server key.

services/core/ForwardHistorian/forwarder/agent.py
```diff
@@ -295,7 +296,11 @@ def historian_setup(self):
                 _log.debug(
                     "Setting up to forward to {}".format(destination_vip))
                 event = gevent.event.Event()
-                agent = Agent(address=destination_vip, enable_store=False)
+                agent = Agent(address=destination_vip,
```
craig8: use the build_agent function from volttron.vip.agent.utils pass the same arguments that you do here.

Fix the event wait in that function while you are at it :)

mikeroup:
@craig8 please look at this again

craig8:
Merged in a6c0bdc0a41fcc1f802abec729c0c9dbd5ae07cf

PR #1037, opened 2016-11-10
Auth subsystem, fixes #519

### LINKED ISSUE #519: Call AuthSerivce.get_capabilities directly (rather than using RPC) from PubSub
PubSub and AuthService run in the same thread. Making the call over RPC introduces unnecessary overhead.

zbeech:
Added auth subsystem to cache capabilities for each user ID. Without this cache the `RPC` and `PubSub` subsystems would have to make a RPC call to the `AuthService` for each capability check. (This means an extra RPC call was made for every `publish` or `rpc.call` to a protected topic or method.)

The lookup key for this cache is the `AuthEntry`'s `user_id` field. The `user-id` field is now required and must be unique. If this field is not specified in the `volttron-ctl auth add` command, then a UUID will automatically be used. This new requirements on the `user_id` field bumped the `AuthFile` version to 1.1. The AuthFile will be automatically updated and it is compatible with 1.0.

Another thing worth pointing out: the default value for `AuthFile.add`'s `overwrite` argument is now `False`. This seems to better apply the "Zen of Python": `Errors should never pass silently. Unless explicitly silenced.` This change should only effect platform code (which has been modified accordingly).

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1037)


craig8:
@zbeech can you look into the conflicts of the main.py.

zbeech:
@craig8 I fixed the conflict in main.py

craig8:
@zbeech great thank you!

Issue #1032, opened 2016-11-08
Make key handling more robust in Core


zbeech:
- Handle empty strings same as `None`
- Throw exceptions if key format is invalid

schandrika:
@schandrika  to test if these are valid and fix as needed

PR #1050, opened 2016-11-15
Datamover agent

mikeroup:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1050)


services/core/DataMover/datamover/agent.py
```diff
@@ -0,0 +1,233 @@
...
+                self._last_timeout = self.timestamp()
+                self._target_platform.core.stop()
+                self._target_platform = None
+                self.vip.health.set_status(
+                    STATUS_BAD, "Timout occured")
+
+        def historian_setup(self):
+            _log.debug(
+                "Setting up to forward to {}".format(destination_vip))
+            event = gevent.event.Event()
```
craig8: Use build connection to create connection directly to the target platforms agent.

volttron/platform/agent/base_historian.py
```diff
@@ -319,6 +319,25 @@ def __init__(self,
         self._process_thread.daemon = True  # Don't wait on thread to exit.
         self._process_thread.start()
 
+    @RPC.export
+    def insert(self, records):
+        for r in records:
```
craig8: document why this is here.

services/core/DataMover/tests/test_datamover.py
```diff
@@ -0,0 +1,696 @@
...
+    """
+    Test if devices topic message is getting forwarded to historian running on
+    another instance. Test if topic name substitutions happened.
+    Publish to topic
+    'datalogger/PNNL/BUILDING_1/Device' in volttron_instance1 and
+    query for topic
+    'datalogger/PNNL/BUILDING1_ANON/Device/MixedAirTemperature' in
+    volttron_instance2
+
+    @param publish_agent: Fake agent used to publish messages to bus in
```
craig8: I don't believe this is well formed sphinx code.  As it it is.  It's hard to read without the spaces below the wrapped txts of the params.

mikeroup:
@craig8 please look at this again

services/core/DataMover/tests/test_datamover.py
```diff
@@ -0,0 +1,696 @@
...
+        # generate public private key pair for instance1
+        ks.generate()
+
+        # add public key of instance1 to instance2 auth file
+        authfile = AuthFile(volttron_instance2.volttron_home + "/auth.json")
+        entry = AuthEntry(credentials=ks.public)
+        authfile.add(entry)
+
+        # setup destination address to include keys
+        forwarder_config["destination-vip"] =\
```
craig8: Hmm this won't be working once the revamp of historian is in the code.

mikeroup: Can you elaborate?

craig8: the distination-vip in the forwarder now is supposed to be only an address.  Though this does bring up a point that we should put a version number around the change between the entire address with pub/secret key in the url.

mikeroup: If the keys are in the url the tests will still pass after making the agent change if you can believe that. I'll edit the tests too to match the forwarder.

craig8: yeah that's actually not a bad thing....if the keys are still in there and the tests pass...however we need to test both ways and they need to be mutually exclusive...

Must be all in url or none but serverkey specified.

craig8: Sorry for the datamover we don't need it to be either or, but for the forwarder we do.

mikeroup: If this is something that should be general then it might fall under #1032

craig8: Probably...this same pattern is going to crop up in all of the agents that are connecting to another platform.  I would like to have a function that takes an agent as a reference and an ip address and it will be able to make a connection based on that.

PR #1067, opened 2016-11-28
Pre 4.0.1 develop merge

craig8:
Making sure the 4.0.1 merge doesn't get too far away from the develop before we have to do nasty merges.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1067)


craig8:
Note merged will do so in another request.

PR #1068, opened 2016-11-28
Added more auth-related volttron-ctl commands

zbeech:
The new commands are:
```
add-group           associate a group name with a set of roles
add-role            associate a role name with a set of capabilities
list-groups         show list of group names and their sets of roles
list-known-hosts    list entries from known-hosts file
list-roles          show list of role names and their sets of capabilities
remove-group        disassociate a group name from a set of roles
remove-known-host   remove entry from known-hosts file
remove-role         disassociate a role name from a set of capabilities
update-group        update group to include (or remove) given roles
update-role         update role to include (or remove) given capabilities
```

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1068)


docs/source/core_services/control/AuthenticationCommands.rst
```diff
@@ -26,20 +26,30 @@ All authentication sub-commands can be viewed by entering following command.
 
     subcommands:
         add                 add new authentication record
+        add-group           associate a group name with a set of roles
         add-known-host      add server public key to known-hosts file
+        add-role            associate a role name with a set of capabilities
         keypair             generate CurveMQ keys for encrypting VIP connections
```
craig8: Can we print that keypair to standand out so that we can easily copy/paste it around

volttrontesting/platform/auth_control_test.py
```diff
@@ -137,3 +167,186 @@ def test_auth_remove(volttron_instance_encrypt):
     entries = auth_list_json(platform)
     assert len(entries) > 0
     assert_auth_entries_same(entries[-1], _auth_entry1.__dict__)
+
+
```
craig8: Can you add a test that tests the restriction using the roles?  Or is what we put on a method only going to be at the capability level?

docs/source/core_services/control/AuthenticationCommands.rst
zbeech: @craig8 Are you referring to the `keypair` command? It does print the public key to stdout (but not the secret key).

volttrontesting/platform/auth_control_test.py
zbeech: @craig8 At least for now, we can only decorate RPC methods with capabilities. Roles and groups just help organize (group) sets of capabilities.

craig8: So for now it's just so that the user can organize their own intent on how the instance should operate.

PR #1090, opened 2016-12-06
add example for actuation via configstore update

mikeroup:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1090)

### DIFF
--- a/docs/source/supporting/examples/ConfigActuation.rst
+++ b/docs/source/supporting/examples/ConfigActuation.rst
@@ -1,7 +1,7 @@
 .. _ConfigActuation:
 
 =========================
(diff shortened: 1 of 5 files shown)


docs/source/supporting/examples/ConfigActuation.rst
```diff
@@ -0,0 +1,23 @@
+.. _ConfigActuation:
+
+=========================
+Config Actuation Example
```
craig8: I talked with @jhaack we are going with the single line under of = signs rather than both below and above for the title.  Please update this throughout your documentation.

PR #1093, opened 2016-12-08
Added global override implementation, pytests and specification doc

shwethanidd:
1. Created spec for global override
2. Updated master driver to support global override
3. Added associated pytests

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1093)

### DIFF
--- a/docs/source/specifications/driver-override.rst
+++ b/docs/source/specifications/driver-override.rst
@@ -1,0 +1,46 @@
+.. _DriverOverride:
+
+Driver Override Specification
+==============================
+This document describes the specification for the global override feature.
+By default, every user is allowed write access to the devices by the master driver. The override feature will allow the user (for example, building administrator) to override this default behavior and enable the user to lock the write access on the devices for specified duration of time .
+
+Functional Capabilities
+-----------------------------
+
+1. User shall be able to specify the following when turning on the override behavior on the devices.
+
+    * Override pattern, for example,
+
+         If pattern is campus/building1/* - Override condition is turned on for all the devices under campus/building1
+
+         If pattern is campus/building1/ahu1 - Override condition is turned on for only campus/building1/ahu1
+
+    * Time duration over which override behavior is applicable.
+
+    * Optional revert-to-fail-safe-state flag. If the flag is set, master driver shall set all the set points falling under the override condition to its default state/value immediately. This is to ensure that the devices are in fail-safe state when the override/lock feature is removed. If the flag is not set, the device state/value is untouched.
+
+    * Optional staggered revert flag. If this flag is set, reverting of devices will be staggered.
+
+2. User shall be able to disable/turn off the override behavior on devices by specifying:
+
+    * Pattern on which the override/lock feature has be disabled. (example: campus/building/*)
+
+3. User shall be able to get a list of all the devices with the override condition set.
+
+4. User shall be able to get a list of all the override patterns that are currently active.
+
+5. User shall be able to clear all the overrides.
+
+Driver RPC Methods
+********************
+set_override_on( pattern, duration=1.0, failsafe_revert=True, staggered_revert=True ) - Turn on override condition on all the devices matching the pattern.
+
+set_override_off( pattern ) - Turn off override condition on all the devices matching the pattern.
+
+get_override_devices( ) - Get a list of all the devices with override condition.
+
+get_override_patterns( ) - Get a list of override patterns that are currently active.
+
+clear_overrides( ) - Clear all the overrides.
+
--- a/docs/source/specifications/driver-override.rst
+++ b/docs/source/specifications/driver-override.rst
@@ -12,11 +12,15 @@ Functional Capabilities
 
     * Override pattern, for example,
 
-         If pattern is campus/building1/* - Override condition is turned on for all the devices under campus/building1
+         If pattern is campus/building1/* - Override condition is turned on for all the devices under campus/building1/.
 
          If pattern is campus/building1/ahu1 - Override condition is turned on for only campus/building1/ahu1
 
+         The pattern matching shall use bash style filename matching semantics.
+
     * Time duration over which override behavior is applicable.
+
+    * Optional indefinite-duration flag for applying override condition indefinitely.
 
     * Optional revert-to-fail-safe-state flag. If the flag is set, master driver shall set all the set points falling under the override condition to its default state/value immediately. This is to ensure that the devices are in fail-safe state when the override/lock feature is removed. If the flag is not set, the device state/value is untouched.
 
@@ -32,11 +36,15 @@ Functional Capabilities
 
 5. User shall be able to clear all the overrides.
 
+6. Any changes to override pattern list shall be  the config store.
+
+7. Whenever a device is newly configured or removed, a check is made to see if it is part of the overridden patterns. If yes, it is added/removed from list of overridden devices.
+
 Driver RPC Methods
 ********************
-set_override_on( pattern, duration=1.0, failsafe_revert=True, staggered_revert=True ) - Turn on override condition on all the devices matching the pattern.
+set_override_on( pattern, duration=1.0, indefinite_duration=False, failsafe_revert=True, staggered_revert=True ) - Turn on override condition on all the devices matching the pattern. The time duration for the override condition has to be in seconds.
 
-set_override_off( pattern ) - Turn off override condition on all the devices matching the pattern.
+set_override_off( pattern ) - Turn off override condition on all the devices matching the pattern. The specified pattern will be removed from the override patterns list. All the devices falling under the given pattern will be removed from the list of overridden devices.
 
 get_override_devices( ) - Get a list of all the devices with override condition.
 
--- a/docs/source/specifications/driver-override.rst
+++ b/docs/source/specifications/driver-override.rst
@@ -3,7 +3,7 @@
 Driver Override Specification
 ==============================
 This document describes the specification for the global override feature.
-By default, every user is allowed write access to the devices by the master driver. The override feature will allow the user (for example, building administrator) to override this default behavior and enable the user to lock the write access on the devices for specified duration of time .
+By default, every user is allowed write access to the devices by the master driver. The override feature will allow the user (for example, building administrator) to override this default behavior and enable the user to lock the write access on the devices for a specified duration of time or indefinitely.
 
 Functional Capabilities
 -----------------------------
@@ -18,9 +18,7 @@ Functional Capabilities
 
          The pattern matching shall use bash style filename matching semantics.
 
-    * Time duration over which override behavior is applicable.
-
-    * Optional indefinite-duration flag for applying override condition indefinitely.
+    * Time duration over which override behavior is applicable. If the time duration is negative, then override condition is applied indefinitely.
 
     * Optional revert-to-fail-safe-state flag. If the flag is set, master driver shall set all the set points falling under the override condition to its default state/value immediately. This is to ensure that the devices are in fail-safe state when the override/lock feature is removed. If the flag is not set, the device state/value is untouched.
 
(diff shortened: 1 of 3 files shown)


docs/source/specifications/driver-override.rst
```diff
@@ -0,0 +1,46 @@
...
+By default, every user is allowed write access to the devices by the master driver. The override feature will allow the user (for example, building administrator) to override this default behavior and enable the user to lock the write access on the devices for specified duration of time .
+
+Functional Capabilities
+-----------------------------
+
+1. User shall be able to specify the following when turning on the override behavior on the devices.
+
+    * Override pattern, for example,
+
+         If pattern is campus/building1/* - Override condition is turned on for all the devices under campus/building1
```
kmonson: Should read:

"all the devices under campus/building1/" (trailing slash).

docs/source/specifications/driver-override.rst
```diff
@@ -0,0 +1,46 @@
...
+
+3. User shall be able to get a list of all the devices with the override condition set.
+
+4. User shall be able to get a list of all the override patterns that are currently active.
+
+5. User shall be able to clear all the overrides.
+
+Driver RPC Methods
+********************
+set_override_on( pattern, duration=1.0, failsafe_revert=True, staggered_revert=True ) - Turn on override condition on all the devices matching the pattern.
```
kmonson: Probably should mention that here that duration is in seconds.

kmonson: Should we have an indefinite duration option?

docs/source/specifications/driver-override.rst
kmonson: We might also mention that pattern matching use bash style filename matching semantics so that it is clear that it's not regex.

kmonson:
I noticed two significant omissions from the spec and the override itself: 

It needs to save the override pattern list to the config store whenever it changes. Should be easy to do.(Was not in the spec, I had written it down and then forgot about it. Sorry about that.)

It does not appear to check newly configured devices to see if they are already overridden.

services/core/MasterDriverAgent/master_driver/agent.py
```diff
@@ -63,15 +63,20 @@
...
 import bisect
+import fnmatch
 
 from driver_locks import configure_socket_lock, configure_publish_lock
 
 utils.setup_logging()
 _log = logging.getLogger(__name__)
 __version__ = '2.0'
 
+class OverrideError(StandardError):
```
kmonson: This should inherit from DriverInterfaceError.

docs/source/specifications/driver-override.rst
shwethanidd: I've added indefinite_duration flag now. I believe this is adding too many checks. Maybe we need to go over it again.

docs/source/specifications/driver-override.rst
```diff
@@ -0,0 +1,46 @@
...
+
+set_override_off( pattern ) - Turn off override condition on all the devices matching the pattern.
```
shwethanidd: Done

docs/source/specifications/driver-override.rst
shwethanidd: Done

docs/source/specifications/driver-override.rst
```diff
@@ -0,0 +1,54 @@
+.. _DriverOverride:
+
+Driver Override Specification
+==============================
+This document describes the specification for the global override feature.
+By default, every user is allowed write access to the devices by the master driver. The override feature will allow the user (for example, building administrator) to override this default behavior and enable the user to lock the write access on the devices for specified duration of time .
```
craig8: If single time period it should be for "a" specified duration.  Is there a way to "lock it indefinately?"

services/core/MasterDriverAgent/master_driver/agent.py
```diff
@@ -400,12 +427,166 @@ def heart_beat(self):
...
+                evt[0].cancel()
+        self._override_interval_events.clear()
+        self._override_devices.clear()
+        self._override_patterns.clear()
+        self.vip.config.set("override_patterns", {})
+
+    # Get a list of all the devices with override condition.
+    @RPC.export
+    def get_override_patterns(self):
+        return list(self._override_patterns)
```
craig8: Add documentation

services/core/MasterDriverAgent/master_driver/agent.py
```diff
@@ -400,12 +427,166 @@ def heart_beat(self):
...
+
+    # Get a list of all the devices with override condition.
+    @RPC.export
+    def get_override_devices(self):
+        return list(self._override_devices)
+
+    # Clear all overrides
+    @RPC.export
+    def clear_overrides(self):
+        # Cancel all pending override timer events
```
craig8: Add documenatation

services/core/MasterDriverAgent/master_driver/agent.py
```diff
@@ -400,12 +427,166 @@ def heart_beat(self):
...
+                    patterns[pat] = 'Indefinite'
+                else:
+                    evt, end_time = self._override_interval_events[pat]
+                    patterns[pat] = end_time.isoformat()
+            self.vip.config.set("override_patterns", jsonapi.dumps(patterns))
+
+    # Turn off override condition on all the entities matching the pattern.
+    @RPC.export
+    def set_override_off(self, pattern):
+        return self._set_override_off(pattern)
```
craig8: Add documentation

services/core/MasterDriverAgent/master_driver/agent.py
craig8: Add documentation

services/core/MasterDriverAgent/master_driver/agent.py
```diff
@@ -400,12 +427,166 @@ def heart_beat(self):
...
+        if path in self._override_devices:
+            raise OverrideError(
+                "Cannot revert point on device {} since global override is set".format(path))
+        else:
+            self.instances[path].revert_point(point_name, **kwargs)
+
     @RPC.export
     def revert_device(self, path, **kwargs):
-        self.instances[path].revert_all(**kwargs)
+        if path in self._override_devices:
```
craig8: Add documentation

services/core/MasterDriverAgent/master_driver/agent.py
```diff
@@ -260,6 +261,17 @@ def configure_main(self, config_name, action, contents):
             except ValueError:
                 pass
 
+        #update override patterns
+        if self._override_patterns is None:
+            try:
+                values = self.vip.config.get("override_patterns")
+                if isinstance(values, dict):
+                    self._override_patterns = set(values.keys())
```
kmonson: What if it's a list or some other type?

services/core/MasterDriverAgent/master_driver/agent.py
```diff
@@ -431,10 +443,10 @@ def revert_device(self, path, **kwargs):
 
     # Turn on override condition on all the devices matching the pattern.
     @RPC.export
-    def set_override_on(self, pattern, duration=1.0, failsafe_revert=True, staggered_revert=False):
-        self._set_override_on(pattern, duration, failsafe_revert, staggered_revert)
+    def set_override_on(self, pattern, duration=1.0, indefinite_duration=False, failsafe_revert=True, staggered_revert=False):
```
kmonson: Instead of adding a new argument could we just use duration <= 0 to signify an infinite duration and make 0 the default?

services/core/MasterDriverAgent/master_driver/agent.py
```diff
@@ -467,66 +490,104 @@ def set_override_off(self, pattern):
...
-        except KeyError:
-            _log.debug("MASTER DRIVER override interval event key error {}".format(topic))
-            pass
-        self._override_interval_events[topic] = self.core.schedule(override_end, self._cancel_override, topic)
+            if config_update:
+                if self._override_interval_events[pat] is None:
+                    patterns[pat] = None
+                else:
+                    evt, end_time = self._override_interval_events[pat]
+                    patterns[pat] = end_time.isoformat()
```
kmonson: Everywhere a datetime object is converted into a string please use volttron.platform.agent.utils.format_timestamp.

services/core/MasterDriverAgent/master_driver/agent.py
```diff
@@ -400,12 +427,166 @@ def heart_beat(self):
...
+                "Cannot revert device {} since global override is set".format(path))
+        else:
+            self.instances[path].revert_all(**kwargs)
 
+    # Turn on override condition on all the devices matching the pattern.
+    @RPC.export
+    def set_override_on(self, pattern, duration=1.0, indefinite_duration=False, failsafe_revert=True, staggered_revert=False):
+        self._set_override_on(pattern, duration, indefinite_duration, failsafe_revert, staggered_revert)
+
+    def _set_override_on(self, pattern, duration=1.0, indefinite_duration=False, failsafe_revert=True, staggered_revert=False):
```
kmonson: Could we use duration <= 0 to indicate infinite duration and make that the default?

services/core/MasterDriverAgent/master_driver/agent.py
```diff
@@ -400,12 +427,166 @@ def heart_beat(self):
...
+        config_update = False
+
+        # If pattern exactly matches
+        if pattern in self._override_patterns:
+            self._override_patterns.discard(pattern)
+            # Cancel any pending override events
+            self._cancel_override_events(pattern)
+            config_update = True
+        else:
+            _log.debug("Pattern did not match!")
```
kmonson: If this happens I think it would make more sense to raise an exception to inform the caller that something went wrong. Nothing after this needs to happen if a matching pattern was not found.

services/core/MasterDriverAgent/master_driver/agent.py
```diff
@@ -400,12 +427,166 @@ def heart_beat(self):
...
+            evt = self._override_interval_events.pop(pattern, None)
+            if evt is not None:
+                evt[0].cancel()
+
+    def _cancel_override(self, topic):
+        self._set_override_off(topic)
+
+    def _update_override_state(self, device, state):
+        device = device.lower()
+
```
kmonson: In the case where state == 'remove' we should not do any comparisons to the override patterns and just remove the device from self._override_devices if it's in there.

services/core/MasterDriverAgent/master_driver/agent.py
```diff
@@ -400,12 +427,166 @@ def heart_beat(self):
...
+        else:
+            self.instances[path].revert_all(**kwargs)
 
+    # Turn on override condition on all the devices matching the pattern.
+    @RPC.export
+    def set_override_on(self, pattern, duration=1.0, indefinite_duration=False, failsafe_revert=True, staggered_revert=False):
+        self._set_override_on(pattern, duration, indefinite_duration, failsafe_revert, staggered_revert)
+
+    def _set_override_on(self, pattern, duration=1.0, indefinite_duration=False, failsafe_revert=True, staggered_revert=False):
+        revert_greenlets = []
```
kmonson: Why are we putting the greenlets in an array if we never do anything with that array?

Also we should probably be using self.core.spawn which does some automatic housekeeping for us.

docs/source/specifications/driver-override.rst
```diff
@@ -36,13 +34,15 @@ Functional Capabilities
 
 5. User shall be able to clear all the overrides.
 
-6. Any changes to override pattern list shall be  the config store.
+6. Any changes to override patterns list shall be stored in the config store. On startup, list of override patterns and corresponding end times are retrieved from the config store. If the end time is indefinite or greater than current time for any pattern, then override is set on the matching devices for remaining duration of time.
```
craig8: On startup, a list of ...

PR #1094, opened 2016-12-08
Fixes for Actuator Agent performance.

kmonson:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1094)


scripts/scalability-testing/config_builder.py
```diff
@@ -67,12 +67,14 @@
 
 class DeviceConfig(object):
```
craig8: We should probably add some documentation to these classes as they are hopefully going to be used by external entities.

services/core/ActuatorAgent/actuator/agent.py
```diff
@@ -712,33 +716,51 @@ def _update_device_state_and_schedule(self, now):
...
             # This won't hurt anything if we are canceling ourselves.
             self._update_event.cancel()
+        self._update_event_time = new_update_event_time
         self._update_event = self.core.schedule(new_update_event_time,
                                                 self._update_schedule_state,
                                                 new_update_event_time)
 
-    def _get_ajusted_next_event_time(self, now, next_event_time):
+    def _get_ajusted_next_event_time(self, now, next_event_time, previously_scheduled_time):
         _log.debug("_get_adjusted_next_event_time")
```
craig8: Consider changing so that you spell the word adjusted correctly in the function name.

kmonson: I'll get those in.

PR #1096, opened 2016-12-08
add configstore update script

mikeroup:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1096)

### DIFF
--- a/examples/ConfigActuation/update_config.py
+++ b/examples/ConfigActuation/update_config.py
@@ -57,9 +57,23 @@
 # }}}
 
 import argparse
-import subprocess
-import tempfile
 import json
+
+from volttron.platform import get_address
+from volttron.platform.agent.known_identities import CONFIGURATION_STORE
+from volttron.platform.keystore import KeyStore, KnownHostsStore
+from volttron.platform.vip.agent.utils import build_agent
+
+def get_keys():
+    """Gets keys from keystore and known-hosts store"""
+    hosts = KnownHostsStore()
+    serverkey = hosts.serverkey(get_address())
+    key_store = KeyStore()
+    publickey = key_store.public
+    secretkey = key_store.secret
+    return {'publickey': publickey,
+            'secretkey': secretkey,
+            'serverkey': serverkey}
 
 def main():
     parser = argparse.ArgumentParser()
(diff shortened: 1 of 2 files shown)


kmonson:
Wouldn't this be simpler with a connection agent using the manage interface? You wouldn't have to create a temporary file.

mikeroup:
It would be pretty similar. We can make the change if you think it would add value.

kmonson:
I would do it.

examples/ConfigActuation/update_config.py
```diff
@@ -0,0 +1,124 @@
...
+    serverkey = hosts.serverkey(get_address())
+    key_store = KeyStore()
+    publickey = key_store.public
+    secretkey = key_store.secret
+    return {'publickey': publickey,
+            'secretkey': secretkey,
+            'serverkey': serverkey}
+
+def main():
+    parser = argparse.ArgumentParser()
```
craig8: Add helper documentation for these

The argument vip_id should be vip_identity.

examples/ConfigActuation/update_config.py
```diff
@@ -0,0 +1,124 @@
...
+# Battelle Memorial Institute. The views and opinions of authors
+# expressed herein do not necessarily state or reflect those of the
+# United States Government or any agency thereof.
+#
+# PACIFIC NORTHWEST NATIONAL LABORATORY
+# operated by BATTELLE for the UNITED STATES DEPARTMENT OF ENERGY
+# under Contract DE-AC05-76RL01830
+
+# }}}
+
```
craig8: Add a description of exactly what this script is going to do for the.  Will it output for standard out etc.  This will be for both read the docs and whoever is looking at the code.  Add references to the config store documentation etc.

examples/ConfigActuation/update_config.py
```diff
@@ -0,0 +1,124 @@
...
+import argparse
+import json
+
+from volttron.platform import get_address
+from volttron.platform.agent.known_identities import CONFIGURATION_STORE
+from volttron.platform.keystore import KeyStore, KnownHostsStore
+from volttron.platform.vip.agent.utils import build_agent
+
+def get_keys():
+    """Gets keys from keystore and known-hosts store"""
```
craig8: Add return param and """ should have its own line.

examples/ConfigActuation/update_config.py
```diff
@@ -0,0 +1,124 @@
...
+
+    try:
+        value = float(args.value)
+    except ValueError:
+        value = args.value
+
+
+    agent = build_agent(**get_keys())
+
+    files = agent.vip.rpc.call(CONFIGURATION_STORE,
```
craig8: Consider using build_connection instead as then you won't need to specify the get(timeout) and you can pass the peer as a argument.  and you can just call "call"

examples/ConfigActuation/update_config.py
craig8: Be more specific here.  From the code in main I am able to deduce that we are updating a leaf in the json tree with a float value.  

Am I not reading this right?

examples/ConfigActuation/update_config.py
```diff
@@ -0,0 +1,138 @@
...
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('vip_identity',
+                        help='VIP Identity whose configstore we want to access')
+    parser.add_argument('filename',
+                        help='Name of the configstore file to edit')
+    parser.add_argument('key',
+                        help='Key to add or edit in the config')
+    parser.add_argument('value',
+                        help='Value that the key will assume')
```
craig8: Value that the key will assume...is this a float, int string double? json string?  Why does it have to be a float?

craig8:
Perhaps rename the script instead of update_config to update_config_key.

PR #1100, opened 2016-12-13
added campus to example command

acedrew:
campus is a required field for device hierarchy, non-standard example was confusing

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1100)


craig8:
Thanks @acedrew

PR #1110, opened 2017-01-03
Rainforest driver

mikeroup:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1110)

### DIFF
--- a/services/core/MasterDriverAgent/tests/test_eagle.py
+++ b/services/core/MasterDriverAgent/tests/test_eagle.py
@@ -1,0 +1,309 @@
+# -*- coding: utf-8 -*- {{{
+# vim: set fenc=utf-8 ft=python sw=4 ts=4 sts=4 et:
+
+# Copyright (c) 2016, Battelle Memorial Institute
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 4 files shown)


services/core/MasterDriverAgent/tests/test_eagle.py
```diff
@@ -0,0 +1,309 @@
...
+        data = None
+
+    start_response('200 OK', [('Content-Type', 'text/html')])
+    return response_dict[data]
+
+    return handle
+
+
+@pytest.fixture(scope='module')
+def agent(request, volttron_instance1):
```
craig8: There is a new method that yield the results rather than using a finalizer so that you can just yield and then cleanup afterwards without having to do a separate function etc.  

I would like us to start moving toward this model as it seems to me as if it's cleaner code.

See https://github.com/VOLTTRON/volttron/blob/develop/services/core/VolttronCentral/tests/test_vc_discovery.py#L18 for an example.

mikeroup: I like this

craig8: Because you aren't passing parameters you don't need the request argument in the function header as well.

PR #1125, opened 2017-01-05
Feature/forwarder timing

kmonson:
I will write tests for the new parse_timestamp_string funtionality later. It works for every case it's actually used for.

Yes, I left the commented out Memory database in base_historian.py. It's mostly finished and I don't want to lose this work.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1125)


kmonson:
This has tests failing on my repo. I'll have to look into it on Monday.

kmonson:
Let's close this until I've got it working.

PR #1133, opened 2017-01-11
add get_multiple_points to actuator and master driver

mikeroup:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1133)


services/core/MasterDriverAgent/master_driver/interfaces/__init__.py
```diff
@@ -396,6 +396,20 @@ def revert_point(self, point_name, **kwargs):
         :param kwargs: Any interface specific parameters.
         """
 
+    def get_multiple_points(self, path, point_names, **kwargs):
+        results = {}
```
craig8: Please document the interface

kmonson:
Make sure there are no other places in the documentation we need to mention this new RPC method and then ship it.

PR #1134, opened 2017-01-12
Added timeout to device config for BACnet Proxy Agent

### LINKED ISSUE #1136: Long running BACnet scrapes timeout in the master driver.
When scraping a large number of points from a slower/older BACnet device, the BACnet proxy can still be scraping when the hardcoded timeout from the master driver expires. This is set to 10 seconds and is not changeable by config. 

This problem was experienced scraping ~1400 points from a JACE that would only support 30 points per request. It is not dependent on scrape interval.

acedrew:
The timeout for how long master driver would wait for the BACnet proxy to scrape requested points was hardcoded to 10 seconds, on large scrapes with slower devices, this timeout could be reached. The timeout is now exposed for each device config. This resolves  #1136 

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1134)

### DIFF
--- a/docs/source/core_services/drivers/BACnet-Proxy-Agent.rst
+++ b/docs/source/core_services/drivers/BACnet-Proxy-Agent.rst
@@ -148,12 +148,10 @@ a device on the first network
 
     {
         "driver_config": {"device_address": "1002:12",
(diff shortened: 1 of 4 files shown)


kmonson:
Looks good to me.

PR #1218, opened 2017-03-01
fixed incorrect source reference, all 'log' sources have already been…

acedrew:
… converted to 'datalogger' when this code runs. Also removed extraneous mongo references in the docstrings, left mongo functions that were commented out in place to stub future work on aggregations.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1218)

### DIFF
--- a/services/core/CrateHistorian/crate_historian/historian.py
+++ b/services/core/CrateHistorian/crate_historian/historian.py
@@ -167,7 +167,7 @@ class CrateHistorian(BaseHistorian):
             else:
                 table = 'device'
 
-        if source == 'log':
+        if source == 'datalogger':
             if db_datatype == 'string':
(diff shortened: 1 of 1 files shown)


services/core/CrateHistorian/crate_historian/historian.py
```diff
@@ -401,7 +401,7 @@ def _build_single_topic_query(self, start, end, agg_type, agg_period, skip,
     def query_historian(self, topic, start=None, end=None, agg_type=None,
                         agg_period=None, skip=0, count=None,
                         order="FIRST_TO_LAST"):
-        """ Returns the results of the query from the mongo database.
+        """ Returns the results of the query from the create database.
```
craig8: Probably want crate here not crate

PR #1230, opened 2017-03-04
created prometheus scrape agent and modified platform.web to support … #inprogress

acedrew:
This agent is probably only interesting for a few folks, but the modifications for platform.web allow non json responses to be delivered by agents. This is not a robust or long term solution, but at least starts the conversation about what needs to be done to support other types of dynamic http content besides json.

I would promote endpoint interface must return a dict or tuple, and that a custom content type can be submitted, I may add that myself in the near future but just wanted to get this out for others to see in mean time.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1230)


acedrew:
Also, apologize for the lack of commit trail here, I developed this on my local machine with a proper commit log, but wasn't able to get volttron to run on my machine, so had to move the project over to a test vm.

Issue #1240, opened 2017-03-10
Mongodb query fails to with naive timestamp


schandrika:
query api fails when the timestamp passed as start/end parameter does not have a explicit timezone information

schandrika:
base historian assumes UTC timezone when inserting, so updated code to default to UTC in query API as well.

PR #1241, opened 2017-03-10
Historian fixes

schandrika:
Rollup script updated based on topic name pattern.
Fixed mongo historian for initial rollup based on timestamp
Fixed issue #1240

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1241)


services/core/MongodbHistorian/mongodb/historian.py
```diff
@@ -345,10 +359,12 @@ def periodic_rollup(self):
                 self.DAILY_COLLECTION, bulk_publish_day, day_ids, db)
 
     def get_last_updated_data(self, db, collection):
+        id = ""
         cursor = db[collection].find({}).sort(
             "last_updated_data", pymongo.DESCENDING).limit(1)
         for row in cursor:
-            return row['last_updated_data']
+            id = row.get('last_updated_data')
```
craig8: Though I think this will do what you want because you are using limit(1) isn't there a find_single or find_one that you can use without having to deal with the loop here?

PR #1257, opened 2017-03-24
New upgrade script

craig8:
Adds a new install-agent script that allows us to very easily install agents in one go with the use of parameters to the script rather than environmental variables.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1257)


craig8:
OBE thanks to #1268

PR #1258, opened 2017-03-25
Develop

rob-calvert:
Here's my submission of Kisensum's new Message Debugger feature. Along with the feature implementation, I've included a user guide and a pytest regression test module. I've put it through a good bit of user testing, but there are still a couple of loose ends (of course!), which I've flagged with @todo's in the code. I intend to keep working on it next week, but I wanted to get the feature out there now so that others can start to look at it. -- Rob Calvert, Kisensum

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1258)


rob-calvert:
This request will be re-submitted.

PR #1285, opened 2017-04-08
fixed error class to catch

### LINKED ISSUE #1284: proxy_grab_bacnet_config.py ValueError not handled
A format string passed the wrong type raises a ValueError, only TypeErrors were being handled, causing the script to fail for devices that did not fit the ":.6g" format string.

acedrew:
Fixes #1284 and cleans up leading whitespace. Maybe we should handle those cases when a string gets passed to "resolution" differently? Currently this will just fail to add the resolution to the notes on the point by excepting and passing.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1285)


kmonson:
I've already fixed this in another commit. This got lost in the shuffle.

craig8:
Really @kmonson  you fixed your formatting trouble.

PR #1286, opened 2017-04-08
PrometheusScrapeAgent

acedrew:
This agent is probably only interesting for a few folks, but the modifications for platform.web allow non json responses to be delivered by agents. This is not a robust or long term solution, but at least starts the conversation about what needs to be done to support other types of dynamic http content besides json.

I would promote endpoint interface must return a dict or tuple, and that a custom content type can be submitted, I may add that myself in the near future but just wanted to get this out for others to see in mean time.

This will dump JSON which prometheus won't be able to understand. I modified my platform.web module to detect when a text property was in the json and just return that as plaintext to make it work, waiting on core team to release suitable API to support this type of use.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1286)


services/contrib/PrometheusScrapeAgent/prometheus_scrape/agent.py
```diff
@@ -75,7 +75,7 @@ def scrape(self, env, data):
             for topic in delete_topics:
                 del self._cache[device][topic]
 
-        return {'text': result}
+        return {'content': result, 'content_type': 'text/plain'}
```
craig8: Change this to 
```
return result, [('Content-Type', 'text/html')]
or
return result, [('Content-Type', 'text/plain')]
````

**Comment by acedrew** [comment]
@craig8 updated

File: services/contrib/PrometheusScrapeAgent/config [hunk 2]
```diff
@@ -0,0 +1,6 @@
+{
+    cache_timeout: 660,
+    tag_delimiter_re: "\s+|:|_|\.|/"
```
craig8: Can you add documentation to what this is actually for?  Our preprocessor strips out # comments so explain the tag delimiter argument.

services/contrib/PrometheusScrapeAgent/prometheus_scrape/agent.py
```diff
@@ -0,0 +1,208 @@
...
+                       "{message_string}".format(topic=topic,
+                                                 message_string=message[0]))
+            raise e
+
+        except IndexError as e:
+            _log.error("message for {topic} missing message string".format(
+                topic=topic))
+            raise e
+
+    def _capture_log_data(self, peer, sender, bus, topic, headers, message):
```
craig8: Why is this function here as it's not used in subscriptions?

acedrew: It will be used, anything that can be cast as a float can also be presented in the scrape page.

acedrew: I added the docs into services/contrib/PrometheusScrapeAgent/docs/README.rst

craig8: On line 38 you subscribe to the driver, but you don't subscribe to the datalogger topics, unless I am missing something in the code?

acedrew: Sorry, I wasn't clear, it will be used in the future. I can take a crack at that now, but I think I will put a flag to disable or enable datalogger collection, as I'm not very interested in that data. I use external telemetry to collect system metrics from the devices the platforms are provisioned on.

craig8: Ahh...what we did with the historians is take the topic list as a configuration and then subscribe over those.

Granted that has downfalls to, but then you can subscribe to an arbitrary topic just as easy as one of the "standard" ones.

PR #1287, opened 2017-04-08
Proposed new web subsystem callback interface to support multiple content types from registered dynamic routes

acedrew:
This will allow agents to specify the exact content type and bypass the default json response structure when needed.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1287)


craig8:
See issue #1298

Issue #1223, opened 2017-03-02
ZMQError: Too many open files


acedrew:
When VC is unavailable for any reason, and the collectors are unable to send data for an extended period of time, once the VC has been restored, this error is in the log from the collector and no data is sent to VC, even if vcp is able to connect.

acedrew:
Still an issue even with latest version of VC and VCP with single socket communication. Happened today with a restart of VC

Issue #1039, opened 2016-11-11
too many sockets open in forwarder


craig8:
````
2016-11-10 02:40:25,359 (forwarderagent-3.6 2892) <stderr> ERROR: Exception AttributeError: "'NoneType' object has no attribute 'set'" in <bound method Socket.__del__ of <volttron.platform.vip.green.Socket object at 0x7eff88a17ae0>> igno
red
2016-11-10 02:40:25,438 (forwarderagent-3.6 2892) <stderr> ERROR: Traceback (most recent call last):
2016-11-10 02:40:25,439 (forwarderagent-3.6 2892) <stderr> ERROR:   File "/home/volttron/volttron-develop/env/local/lib/python2.7/site-packages/gevent/greenlet.py", line 534, in run
2016-11-10 02:40:25,525 (forwarderagent-3.6 2892) <stderr> ERROR:     result = self._run(*self.args, **self.kwargs)
2016-11-10 02:40:25,526 (forwarderagent-3.6 2892) <stderr> ERROR:   File "/home/volttron/volttron-develop/volttron/platform/vip/agent/core.py", line 283, in run
2016-11-10 02:40:25,556 (forwarderagent-3.6 2892) <stderr> ERROR:     looper.next()
2016-11-10 02:40:25,557 (forwarderagent-3.6 2892) <stderr> ERROR:   File "/home/volttron/volttron-develop/volttron/platform/vip/agent/core.py", line 562, in loop
2016-11-10 02:40:25,557 (forwarderagent-3.6 2892) <stderr> ERROR:     self.socket = vip.Socket(self.context)
2016-11-10 02:40:25,557 (forwarderagent-3.6 2892) <stderr> ERROR:   File "/home/volttron/volttron-develop/volttron/platform/vip/socket.py", line 303, in __new__
2016-11-10 02:40:25,572 (forwarderagent-3.6 2892) <stderr> ERROR:     return base.__new__(cls, context, socket_type)
2016-11-10 02:40:25,573 (forwarderagent-3.6 2892) <stderr> ERROR:   File "zmq/backend/cython/socket.pyx", line 281, in zmq.backend.cython.socket.Socket.__cinit__ (zmq/backend/cython/socket.c:3220)
2016-11-10 02:40:25,592 (forwarderagent-3.6 2892) <stderr> ERROR: ZMQError: Too many open files
[...]

PR #1320, opened 2017-04-26
To fix issue 1223, 1039. 

shwethanidd:
When an agent is created, two sockets are created. 
1. VIP socket to send and receive messages over the message bus. 
2. Monitor socket to monitor the VIP socket connection. 
So when the agent is stopped, these 2 sockets are not getting closed. This is leading to build up of open sockets over time (in the case of forward historian). 
Fix: Closed the open sockets when an agent is stopped or VIP socket is disconnected.
Catch ENOTSOCK ZMQError in the forward historian and stop the target platform agent.
Added a disconnect handler in RPC subsystems.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1320)


jhaack:
Listener agent doesn't work in this pull request.

craig8:
Note Add # in front of the numbers or they won't be referenced by the individual tickets
#1223, #1039

craig8:
Because you merged from develop this pull request is tainted :(

PR #1328, opened 2017-04-28
Fixed #1327 During vcfg setup the password is now not echoed to stdout.

### LINKED ISSUE #1327: volttron-cfg should not echo password for volttron central

craig8:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1328)


jhaack:
Since you can't see the password anymore, might want to copy the password confirmation pattern (the user enters the password again and do a match check)

PR #1335, opened 2017-04-29
Added SelectiveHistorian agent

kmorri09:
The SelectiveHistorian agent is the exact same as the base historian from services/core/, except I have added a few lines of code to be able to filter the topics that are sent to the database. This is useful when creating dedicated databases for specific devices being scraped.

All new code (from base historian) is marked with an all-caps comment.

Known issue: when the historian is written to less frequently than every 15 seconds, it may time out and stop writing. It's unclear if this is a result of the new lines of code or an issue with the base historian, or a combination.

Contact me (kmorri09@gmail.com) for questions on this agent.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1335)


examples/SelectiveHistorian/sqlhistorian/historian.py
```diff
@@ -0,0 +1,399 @@
...
+                        len(to_publish_list)-topics_blocked)) #MODIFIED TO FILTER TOPICS
+                    self.report_all_handled()
+                else:
+                    msg = 'commit error. rolling back {} values.'
+                    _log.debug(msg.format(len(to_publish_list)))
+                    self.writer.rollback()
+            elif topics_blocked < len(to_publish_list): #MODIFIED TO FILTER TOPICS
+                _log.debug(
+                    'Unable to publish {}'.format(len(to_publish_list)-topics_blocked)) #MODIFIED TO FILTER TOPICS
+        except:
```
duboviy: - To catch any exception use the ```except Exception```, instead of ```except BaseException``` or just ```except```.
- Indicate the most specific exception type in ```except``` block.

For example:
```python
# Bad
try:
    mapping[key]
except Exception:
    ...

# Better
try:
    mapping[key]
except KeyError:
    ...
```

craig8:
@kmorri09 please use a different branch so we can distinguish between the selective historian pull request and a weather agent pull request.

kmorri09:
This agent is based on the SQL historian class. All I have done is added some logic to the SQL historian so that the variables in the config file can be used to sort through which topics should be saved in the database. A more accurate name may be "SelectiveSQLHistorian." I am sure similar logic may be applied to the other historian types, or to the base historian, but since I only deal with SQL this is what I came up with for our solution.

weather agent removed, sorry for that confusion.

Issue #1227, opened 2017-03-04
Documentation cleanup


jhaack:
Cleaning up minor issues in documentation

craig8:
I would say we did enough here for this round of clean up :) for 4.1 release anyways

PR #1348, opened 2017-05-09
Cleaning up the setup index

jhaack:
#1227 Cleaning up setup index

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1348)


craig8:
Note should these be picked into releases/4.1 branch?

PR #1352, opened 2017-05-11
Bacpypes imports fix

davidraker:
The new version of BACpypes has changes to module names which break the import statements in the BACnet helper scripts.  Updating the import statements per http://bacpypes.readthedocs.io/en/stable/migration/migration001.html#localdeviceobject allowed me to successfully run bacnet_scan.py and grab_bacnet_config.py.  I also updated one other place where the same import statements appear, though I have not tested anything there. There is a fourth instance, which was not updated, but it is in a directory marked deprecated.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1352)


acedrew:
Looks good, these were the changes I was going to make :)

PR #1360, opened 2017-05-23
Check on reasoning

craig8:
@ffreckle 
@rob-calvert

The removed comments below are your internal auth cache or the platform auth cache.  The platform auth cache will be updated every time you install an agent.  So with the code right above it, it would be updated 3 times.

If you added --start to the commands that would start the agents as well.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1360)


craig8:
I believe that this has become OBE correct
@ffreckle 
@rob-calvert

PR #1362, opened 2017-05-23
Updated weather agent

maskedpatel:
Readme contains usage information as well as future changes.
New weatheragent is under services/core. Can make changes based on feedback.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1362)


craig8:
If this is supposed to replace the WeatherAgent then why did you lowercase the main container, it makes it very difficult to compare what you modified in this format.

maskedpatel:
Right sorry, the name 'updated weather agent' probably wasn't the best idea. Our goal was to have this be another option to the already existing WeatherAgent, not to replace it. Ours publishes multiple weather reports periodically which may be a useful functionality for many users. (More written about it in the readme)

craig8:
Ahh ok then perhaps I would put this under contrib rather than services core and then you can give your company a plug.  as well as naming it in Camel Case.

PR #1363, opened 2017-05-23
Add example CSV historian.

kmonson:
Example CSV historian agent.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1363)


examples/CSVHistorian/csv_historian/historian.py
```diff
@@ -0,0 +1,142 @@
...
+
+            row["source"] = record["source"]
+            row["topic"] = record["topic"]
+            row["value"] = record["value"]
+
+            self.csv_dict.writerow(row)
+
+        self.report_all_handled()
+        self.f.flush()
+
```
craig8: please add stubs for the methods that a "real" historian would need to implement. i.e a query_historian function.

PR #1375, opened 2017-06-01
OpenADR interface spec

rob-calvert:
Spec for a new VENAgent that participates in OpenADR demand response events.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1375)


craig8:
thanks @rob-calvert

Issue #1367, opened 2017-05-24
Return Custom Status code in WebAgent response


sheridaj:
Currently, the web agent will only return a 200 OK response for a successful call to an endpoint.  It would be nice to return other status codes defined by the callback function.

Currently, the callback function can return a tuple of length 2.  The first element in the tuple is the data for the response, the second element is a list of tuples, each tuple being a header-name, value.
e.g. `return "My Data", [("Content-Type", "application/sep+xml")]`

Proposed solution: 
Make the first element a dictionary: e.g.  `return {"data": "My Data", "status_code": "200 OK"}, [("Content-Type", "application/sep+xml")]`

However, there are other ways to implement this change.  If the community has any feedback or preference about implementation, it is more than welcome

craig8:
Adding some context we are talking about the method create_response

https://github.com/VOLTTRON/volttron/blob/develop/volttron/platform/web.py#L627

```
def create_response(self, res, start_response): 
```

In the current implementation res can be a dictionary (for json-rpc), a or a list/tuple with the first element the data and the second element a "headers" element to specify return type, or a string which will be turned into json.

So the solution above would turn the first element of the tuple into a dictionary with data and status_code as the keys.

The assumption being that if that first element isn't a dictionary it would just be data with a default 200 OK code?

acedrew:
I don't think it's unreasonable to expect every endpoint to know its appropriate status code. Why not make it just a flat tuple of (status, data, headers)? I think I'm the only one currently affected by changes to this, and I'm happy to adapt to something that will be more consistent long term.

craig8:
I agree, however I would suggest that having to set the status code on all endpoints seems repetitive, especially since most will be a 200 OK (I would assume).  That being said I do like the flat tuple that moves status to the end...so

(data, headers, status) if status is not set then have it be a default '200 OK'

I'll add tests to make sure this backward compatible and we can go from there.

acedrew:
Shouldn't you be able to send a bad status with no content?

ffreckle:
Yes and SEP2 specifies HTTP 201 and 204 with no content.  Perhaps this is a good time to address the implied todo in the code?

>  Note this is specific to volttron central agent and should
>  probably not be at this level of abstraction.

If `res` was a response dictionary with keys: 'data', 'status_code' and 'headers', you could do convenient defaults, like 200, and follow a common convention for returning an HTTP response. Would it be a large effort to modify Volttron Central to use this interface?

[...]

Issue #1234, opened 2017-03-05
No support for gz compression in web subsystem


acedrew:
Some of our platforms are on wireless WAN connections, it would be nice to have gz so that we were using less bandwidth.

craig8:
Links for implementor

https://stackoverflow.com/questions/8506897/how-do-i-gzip-compress-a-string-in-python
https://developer.mozilla.org/en-US/docs/Web/HTTP/Compression
https://en.wikipedia.org/wiki/HTTP_compression

acedrew:
Another reason this is important, modern processors are optimized for compression workloads, and it can be faster to compress data before transmitting than transmitting in general, even without hitting network IO. Windows takes advantage of this and actually compresses data in RAM while programs are running.

craig8:
I am going to suggest not to take #1379 as is, but instead wait for the discussion on #1378 and epic #1382 to handle this.

PR #1379, opened 2017-06-03
Gzip feature branch

acedrew:
This implements gzip encoding in the web subsystem, based on implicit presence of the Content-Encoding header in the registered endpoint response tuple. I don't think this is the right way to do it, and I think this should be solved in a comprehensive way by the rework being discussed in #1367, but this solved my problem (#1234) in the short term, and I wanted to share it for others and to help fuel the discussion around a permanent solution. I'm happy to implement suggestions for this as well.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1379)


craig8:
Closed as the merge is consumed by #1418

Issue #1382, opened 2017-06-04
Refactor web.py to make more generic


craig8:

acedrew:
Epic haha, I like Flask a lot as well, it makes sense to try to emulate, let me know how I can help with this.

acedrew:
Thoughts on this? https://github.com/acedrew/volttron/commit/341bd6cfac050dbd5d0df373c6ab5e51cde39c00

sheridaj:
The code you submitted along with #1388 works great for my purposes. Already forked your repo and am running with it!

ffreckle:
As @sheridaj indicates, we will use the refactor and WebResponse (thank you) but this discussion has me wondering:

- What are are the pros/cons of building our own WSGIApplication within Volttron rather than adopting an open source library like Flask?  Is there a technical reason this is not possible/reasonable?

- What is the argument against hosting the WSGIApplication within a Volttron agent so that requests intended for this agent do not need to travel back and forth across the bus?

I'm not suggesting a change here so much as I am wanting to understand the design philosophy behind these decisions.  I'll contribute the pros/cons that I'm aware of:

Open Source WSGI Library (eg Flask):

Pros: 
- Widely deployed, heavily tested 
- Well documented and understood by developers
- Rich feature set - routing, variable rules, reverse urls, sessions, etc.
- Less new code to write

Cons (needs help):
- Another dependency
- (?) Technical reason why it cannot be used within Volttron
- (?) More complex than is required
- (?) License issue

Agent hosting of WSGI Application (rather than Volttron/MasterWebService):

Pros:
- Avoid message bus traffic that is destined for single agent (eg all SEP2 traffic)
- Enable use of 3rd party library (like Flask) - Could create a abstract WebAgent class 
- Distributes workload, offloading main volttron process

Cons:
- Need to figure out how to manage ports across agents
- Connecting applications need to know correct port
- (?) Security issue

acedrew:
I'm a little nervous about Flask's  Context Globals in a platform as complex as Volttron. If Volttron had been built around Flask from the ground up, it would be a different story, but I'm afraid with multiple agents using the same Flask App, things could get really complicated and make things even more non-deterministic. 

Flask also does not provide what they consider a production ready implementation of a WSGI server, so we would still have to use the main process as the underlying server, or spin up a new, separate HTTP server process. (this could be an option in the future for low power nodes with high levels of HTTP workload)

I am not a Lawyer, but the Licenses seem to be compatible.

[...]

Issue #1378, opened 2017-06-03
Content returned from registered endpoints must be encodable as string.


acedrew:
While the underlying WSGI interface supports a raw bytes interface (indeed that is the specification) the web subsystem expects a serializable string, preventing endpoints from implementing their own encoding.

PR #1388, opened 2017-06-04
Web refactor to support custom headers, binary data, and status codes

acedrew:
This is an example implementation of these features added in a backwards compatible way with agents using the JSON-RPC interface.

This addresses #1378, #1379, #1382, #1367 

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1388)

### DIFF
--- a/services/contrib/PrometheusScrapeAgent/prometheus_scrape/agent.py
+++ b/services/contrib/PrometheusScrapeAgent/prometheus_scrape/agent.py
@@ -37,6 +37,7 @@ class PrometheusScrapeAgent(Agent):
     @Core.receiver("onstart")
     def _starting(self, sender, **kwargs):
         self.vip.web.register_endpoint('/promscrape', self.scrape, "raw")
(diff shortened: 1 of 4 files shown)


craig8:
Looks like you Brooke some tests

acedrew:
working through the tests now

sheridaj:
in create_raw_response in web.py, if res is a tuple of length 1 or 2, the call to start_response will crash because headers are not defined.

acedrew:
Haha, yeah I forgot to go back and fillin/handle default values... That can
be easily resolved but I think @craig8 preferred we build an actual
response object with the appropriate helper functions, which is shown in
skeleton form

acedrew:
@sheridaj is this functional? Obviously we're supposed to provide a content-type header, but I wasn't sure what would be the appropriate one to default to. I also don't think it's appropriate for the create_raw_response method to implement any mime detection or other higher level features, those should be built on top of this method.

Issue #1344, opened 2017-05-04
Handle registration of web endpoints better


craig8:
Currently if a platform hasn't been "web enabled" the following trace will happen without really providing the context to fix the issue.

````
2017-05-04 17:27:20,836 () volttron.platform.web DEBUG: REGISTERING ENDPOINT: /vc/ws
2017-05-04 17:27:20,836 () volttron.platform.vip.agent.subsystems.rpc ERROR: unhandled exception in JSON-RPC method 'register_websocket': 
Traceback (most recent call last):
  File "/opt/volttron/volttron/volttron/platform/vip/agent/subsystems/rpc.py", line 168, in method
    return method(*args, **kwargs)
  File "/opt/volttron/volttron/volttron/platform/web.py", line 497, in register_websocket
    self.appContainer.create_ws_endpoint(endpoint, identity)
AttributeError: 'NoneType' object has no attribute 'create_ws_endpoint'
````

The only way this happens is when the platform hasn't been started with a web address.

acedrew:
What were the steps for recreating this?

PR #1391, opened 2017-06-06
log message when attempting to register websocket without web

acedrew:
This should address #1344, need a good way to test, if agreeable, this could be applied to the rpc exported functions in the web system.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1391)


craig8:
Agreed

craig8:
So does that mean you are going to add it the rpc exported methods?

PR #1405, opened 2017-06-20
Better messges for debugging drivers.

kmonson:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1405)

### DIFF
--- a/examples/ListenerAgent/listener/agent.py
+++ b/examples/ListenerAgent/listener/agent.py
@@ -62,6 +62,8 @@ from datetime import datetime
 import logging
 import sys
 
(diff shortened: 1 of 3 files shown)


fstshrk:
How long do we expect the "Message" to be. Is there any chance that we want to cut it off at a fixed length?

kmonson:
The only difference in the message length is the formatting white space. I guess we could cut it off, but then it's not as useful to a reader later. What is your concern?

kmonson:
And the message format only changes for the ListenerAgent which shouldn't be running all the time anyway.

PR #1421, opened 2017-07-18
Create extract_config_store.py script.

kmonson:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1421)

### DIFF
--- a/scripts/extract_config_store.py
+++ b/scripts/extract_config_store.py
@@ -1,0 +1,132 @@
+# -*- coding: utf-8 -*- {{{
+# vim: set fenc=utf-8 ft=python sw=4 ts=4 sts=4 et:
+
+# Copyright (c) 2016, Battelle Memorial Institute
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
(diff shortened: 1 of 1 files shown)


scripts/extract_config_store.py
```diff
@@ -0,0 +1,132 @@
...
+    event = gevent.event.Event()
+    gevent.spawn(agent.core.run, event)
+    event.wait()
+
+    config_list = agent.vip.rpc.call(CONFIGURATION_STORE,
+                           'manage_list_configs',
+                           config_id).get(timeout=10)
+
+    if not config_list:
+        print "Config store", config_id, "does not exist."
```
craig8: Please use 3.x semantics for print eventually we will move to 3x python

scripts/extract_config_store.py
```diff
@@ -0,0 +1,132 @@
...
+        return
+
+    ensure_dir(output_directory)
+
+    os.chdir(output_directory)
+
+    for config in config_list:
+        print "Retrieving configuration", config
```
craig8: Please use 3.x semantics for print eventually we will move to 3x python

Issue #1297, opened 2017-04-17
when pycharm developing AGENT_CONFIG and VOLTTRON_HOME shouldn't be required


craig8:
AGENT_CONFIG should be added as default {} file that is generated automatically in the pycharm-launch script.

schandrika:
VOLTTRON_HOME is defaulted. Still expects either AGENT_CONFIG to be set or looks for a file called 'config' in agent directory

PR #1426, opened 2017-08-03
Fixes #1022 auto adding auth if agent id doesn't exist

### LINKED ISSUE #1022: Auth gets in the way when running from command line
Starting an agent through eclipse or pycharm or from the command line is problematic because each individual agent has to be in the auth.json file.  I recommend we update pycharm-launch.py and pydev-launch.py to automatically add those agents that aren't using the default platform keystore to make sure their runtime agents are added to the auth.json file.

Of course adding /.*/ credential in the auth.json file is a way of bypassing this, however NOT to be used in production!

craig8:
This pull request addresses #1022 (at least for pycharm) and #1297 for VOLTTRON_HOME.  The AGENT_CONFIG should default to config if config is present.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1426)

### DIFF
--- a/scripts/pycharm-launch.py
+++ b/scripts/pycharm-launch.py
@@ -20,6 +20,8 @@ import argparse
 import sys
 import os
 import runpy
+import subprocess
+import json
 
 __author__ = 'Craig Allwardt<craig.allwardt@pnnl.gov>'
 __version__ = '1.0.0'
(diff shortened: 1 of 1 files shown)


jhaack:
Looks good. Much more informative.

PR #1432, opened 2017-08-07
vcp respects config store and instances name

craig8:
The guid for instances is now being replaced with an id that relies on the instance name to be set.

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1432)


craig8:
Closing as more work to do.

PR #1468, opened 2017-09-06
Develop2

cyrus19901:
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1468)


cyrus19901:
Removing the VOLTTRON log and adding hyperlinks

Issue #1470, opened 2017-09-09
Change json.py to use zmq.utils.jsonapi


shwethanidd:
To enable json.py to use zmq,utils.jsonapi for now and swap it with a different json utility if needed later.

PR #1471, opened 2017-09-09
wrap json library in our own so we can easily test different libraries

shwethanidd:
Issue #1470 Use zmq.utils.jsonapi instead of ujson

---
This change is [Reviewable](https://reviewable.io/reviews/volttron/volttron/1471)


craig8:
Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/volttron/volttron/1471)*

PR #1866, opened 2018-12-20
Update volttron home references for volttron security assessment

jklarson:
# Description

Changed files to use get_home() and set_home() from volttron.platform.

# How Has This Been Tested?

Tested using existing tests in services and volttrontesting.

# Checklist:

### DIFF
--- a/services/core/OpenADRVenAgent/test/test_ven.py
+++ b/services/core/OpenADRVenAgent/test/test_ven.py
@@ -47,7 +47,7 @@ import requests
 import sqlite3
 import time
 
-from volttron.platform import get_services_core
+from volttron.platform import get_services_core, get_home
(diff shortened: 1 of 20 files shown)


craig8:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/VOLTTRON/volttron/pull/1866'>Refresh</a>.<a href='#crh-update-none'></a>

craig8:
@jklarson so lets look at this and see that there are a heck of a lot of os.path.join(get_home(), 'path') going on here.  If you look at the volttron.platform.__init__ file and see that the get_core_services(path) handles that for us.  I wonder if it would make sense to modify this code to deal with this that way rather than through os.path?

What do you think?

PR #1902, opened 2019-02-08
Update TopicWatcher Agent

### LINKED ISSUE #1899: Topic Watcher alert to remote machine
Allow alerts to be able to be sent directly to a secondary instance if data isn't published to the primary instance.

craig8:
# Description

Updates the TopicWatcher to send alerts to a remote platform if a device doesn't publish in a specified period of time.  This does not change existing behaviour only enhances and allows a remote publishing of the alert.

Fixes #1899 

## Type of change

# How Has This Been Tested?

Tested on PNNL campus deployment

# Checklist:

### DIFF
--- a/services/ops/TopicWatcher/config
+++ b/services/ops/TopicWatcher/config
@@ -1,4 +1,15 @@
 {
+    "publish-settings":
+    {
+        "publish-local": false,
+        "publish-remote": true,
+        "remote":
+        {
+            "serverkey": "Olx7Y7XZvSGmHDppsQKvG7BucOH8vgkRlQGZzzh5nHs",
+            "vip-address": "tcp://127.0.0.1:23916",
+            "identity": "remote.topic_watcher"
+        }
+    },
     "group1": {
         "devices/fakedriver0/all": 10
     },
--- a/services/ops/TopicWatcher/topic_watcher/agent.py
+++ b/services/ops/TopicWatcher/topic_watcher/agent.py
@@ -46,13 +46,14 @@ from volttron.platform.agent.known_identities import PLATFORM_TOPIC_WATCHER
 from volttron.platform.agent import utils
 from volttron.platform.messaging.health import Status, STATUS_BAD
 from volttron.platform.vip.agent import Agent, Core, RPC
+from volttron.platform.vip.agent.utils import build_agent
 from volttron.platform.agent.utils import get_aware_utc_now
 from volttron.platform.scheduling import periodic
 
 utils.setup_logging()
 _log = logging.getLogger(__name__)
 
-__version__ = '1.0'
+__version__ = '2.0'
 
 
 class AlertAgent(Agent):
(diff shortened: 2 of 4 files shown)


services/ops/TopicWatcher/config
```diff
@@ -1,4 +1,15 @@
 {
+    "publish-settings":
+    {
+        "publish-local": false,
+        "publish-remote": true,
+        "remote":
```
schandrika: if there is a "remote" config it could imply that publish-remote is true. So why have a separate config "publish-remote"

services/ops/TopicWatcher/topic_watcher/agent.py
```diff
@@ -61,7 +62,44 @@ def __init__(self, config_path, **kwargs):
         self.config = utils.load_config(config_path)
         self.group_agent = {}
         self._connection = None
-
+        self.publish_settings = self.config.get('publish-settings')
+        self._remote_agent = None
```
schandrika: so we can have only one remote agent?

services/ops/TopicWatcher/topic_watcher/agent.py
```diff
@@ -61,7 +62,44 @@ def __init__(self, config_path, **kwargs):
...
+            self.remote_address = remote.get('vip-address', None)
+
+            if self.publish_remote:
+                assert self.remote_identity
+                assert self.remote_serverkey
+                assert self.remote_address
+
+    @property
+    def remote_agent(self):
+        if self._remote_agent is None:
```
schandrika: If we create the remote agent only once, why not create the agent when reading config and pass the remote agent to AlertGroup?

services/ops/TopicWatcher/config
craig8: So you can turn it on or off regardless if the section is present or not

services/ops/TopicWatcher/topic_watcher/agent.py
craig8: Are you asking if we should allow the sending to more than one instance?

services/ops/TopicWatcher/topic_watcher/agent.py
craig8: because if the agent disconnects or doesn't start up then the lower level group agent doesn't know how to connect.  Also, there are many AgentGroup gevent processes that all share the main agent.

services/ops/TopicWatcher/topic_watcher/agent.py
schandrika: yes

craig8: I don't think so in this case...though that's an interesting thought.  We are just sending it from devices to a vc instance where an emailer will send it off to the correct party.

services/ops/TopicWatcher/topic_watcher/agent.py
schandrika: Reconnect happens if the very first ping to the remote instance fails, but if the connection fails after the first successful alert then it won't reconnect.

craig8: Fixed!

PR #1910, opened 2019-02-18
Rabbitmq certs test

schandrika:
Test cases for ssl cert expiry use cases.


volttron/utils/rmq_mgmt.py
```diff
@@ -783,6 +783,10 @@ def build_connection_param(self, rmq_user, ssl_auth=None, retry_attempt=30, retr
                     ssl_options=ssl_options,
                     credentials=pika.credentials.ExternalCredentials())
             else:
+                # TODO: How is this working? PlainCredentials(rmq_user,
+                # rmq_user) ?? My understanding is that non ssl mode is going to
+                #  be used only for testing - when using plain
+                # credentials all agents use same password.
```
shwethanidd: For non-ssl, username and password is the same.

PR #1911, opened 2019-02-19
Csr requests 

craig8:


volttron/platform/vip/agent/core.py
```diff
@@ -881,17 +887,23 @@ def _build_connection_parameters(self):
...
-                                                         self.instance_name)
+                if self.instance_name == get_platform_instance_name():
+                    param = self.rmq_mgmt.build_agent_connection(self.identity,
+                                                                 self.instance_name)
+                else:
+                    param = self.rmq_mgmt.build_remote_connection_param(self.rmq_user,
             except AttributeError:
                 _log.error("RabbitMQ broker may not be running. Restart the broker first")
                 param = None
+                self.rmq_address,
```
shwethanidd: I think this is not merged properly.

volttron/platform/vip/agent/core.py
```diff
@@ -881,17 +887,23 @@ def _build_connection_parameters(self):
...
+                                                                    True)
         return param
 
     def loop(self, running_event):
-        param = self._build_connection_parameters()
+        if not isinstance(self.rmq_address, pika.ConnectionParameters):
+            self.rmq_address = self._build_connection_parameters()
```
shwethanidd: This method is not merged correctly. The return value of self._build_connection_parameters() is stored self.rmq_address but method still uses previous parameter 'param'

PR #1916, opened 2019-02-28
Csr approval update

craig8:
# Description

This pull request brings an endpoint /admin/api/approve_csr/<common_name> to the server.


volttron/platform/vip/agent/subsystems/auth.py
```diff
@@ -76,30 +79,78 @@ def onsetup(sender, **kwargs):
...
+                    # is installed.
+                    if get_messagebus() == 'rmq':
+                        if not os.path.exists("keystore.json"):
+                            with open("keystore.json", 'w') as fp:
+                                fp.write(json.dumps(KeyStore.generate_keypair_dict()))
+
+                        with open("keystore.json") as fp:
+                            keypair = json.loads(fp.read())
+
+                    return build_agent(agent_class=agent_class,
```
schandrika: Could set it as the variable 'value' so that there is a single return statement at the end.

volttron/platform/web/csr_endpoints.py
```diff
@@ -51,13 +63,11 @@ def _csr_request_new(self, env, data):
...
-        csr_file = self._certs.csr_pending_file(identity)
-        if csr:
-            with open(csr_file, "wb") as fw:
-                fw.write(csr)
 
-        auto_accept = True
+        auto_accept = False
         if auto_accept:
+            if self._certs.cert_exists(identity, True):
+                cert = self._certs.cert(identity, True)
```
schandrika: Line 76 any way create a new signed cert from csr so why load this cert?

craig8: Currently no test cases...I will add to a later pull request.  If you want to test this you will need to vms that can connect to each other.  I have to make documentation for this as well today.  Again another pull request to come.

PR #1926, opened 2019-03-06
Fixes #1925 handles issues with current configuration

### LINKED ISSUE #1925: Update of config doesn't handle current config properly
Related to the fixes of #1924 when the config file exists already there is an error thrown because the config object doesn't have a volttron section.

craig8:
Fixes #1925  Bug in implementing update_config


craig8:
Superseded by #1929

PR #1935, opened 2019-03-09
added ability to add web-ssl-cert and web-ssl-key from vcfg.

schandrika:


volttron/platform/instance_setup.py
```diff
@@ -391,13 +400,87 @@ def vc_config():
...
+                'Enter the SSL certificate private key file:'
+            key_file = prompt_response(prompt, mandatory=True)
+            if is_file_readable(key_file):
+                break
+            else:
+                print("Unable to read file {}".format(key_file))
+        try:
+            if certs.Certs.validate_key_pair(cert_file, key_file):
+                cert_error = False
+                shutil.copy(cert_file, master_web_cert)
```
craig8: I don't think that these should be copied into the masterweb.  This will probably cause a problem with masterweb not being able to connect to rabbitmq as I would doubt that the cert and key have been signed by our root ca.  We should either copy to a new specific location or leave as is with just the filename.

volttron/utils/rmq_setup.py
```diff
@@ -659,14 +659,12 @@ def _create_rabbitmq_config(rmq_config, setup_type):
                     while True:
                         prompt = 'Enter the root CA certificate public key file:'
                         root_public = prompt_response(prompt, mandatory=True)
-                        root_public = os.path.expanduser(root_public)
                         if is_file_readable(root_public):
```
craig8: Why are you not continuing to expanduser (should be expandall)?  root_public unless I am missing something will never get expanded to the full path.  The is_file_readable function does the expansion, but the passed string is not changed and it shouldn't be.

volttron/utils/rmq_setup.py
```diff
@@ -724,6 +722,7 @@ def _create_rabbitmq_config(rmq_config, setup_type):
 
 
 def is_file_readable(file_path):
+    file_path = os.path.expanduser(file_path)
```
craig8: change to expandall

volttron/utils/rmq_setup.py
craig8: https://volttron.readthedocs.io/en/develop/_modules/volttron/platform/config.html#expandall

craig8:
I think we need to separate the the web server from the installation of vc.  See the following 
````
(volttron)osboxes@v2 ~/repos/volttron-rabbitmq $ vcfg

Your VOLTTRON_HOME currently set to: /home/osboxes/.volttron-rabbitmq

Is this the volttron you are attempting to setup?  [Y]: 
What is the external instance ipv4 address? [tcp://127.0.0.1]: 
What is the instance port for the vip address? [22916]: 
What is type of message bus? [zmq]: rmq
Is this instance a volttron central? [N]: 
Will this instance be controlled by volttron central? [Y]: n
Would you like to install a platform historian? [N]: 
Would you like to install a master driver? [N]: 
Would you like to install a listener agent? [N]: 
Finished configuration

You can now start the volttron instance.

If you need to change the instance configuration you can edit
the config file at /home/osboxes/.volttron-rabbitmq/config
````

craig8:
ACK another issue is that the prompt for username/password needs to change to use the new method.  I will look at that in a bit.

volttron/platform/instance_setup.py
schandrika: we are creating this only in case of zmq + https web-bind-address. So it is not connecting to rmq and won't have any self signed root certs. Or are you thinking of a case where someone switches from RMQ to ZMQ and might want to go back to RMQ with the same vhome? I wouldn't recommend them doing it, but guess in theory it could be done

craig8: no I am saying that if the user has any cert key we should be able to use it rather than copying it so if someone has a cert and key in ~/.ssh/ and it has a 400 permission on it, then it shouldn't get spread/copied/moved to the masterweb location

schandrika: Valid point. But what if the path is $HOME/filename or $HOME/Downloads/filename or /tmp/filename. I am just thinking it might get accidentally deleted. Should we print some kind of warning or info that we will be simply pointing to it? Or may be we can print the final generated config file at the end?

craig8: If it accidentally gets deleted then they will get an error the next time volttron is run because they won't have a valid path.  

We can print out a warning saying that if the file isn't chmod 400 then its perhaps in danger just like ssh does if you use an identity file connecting that is group writable.

schandrika:
This PR is for issue #1935

PR #1942, opened 2019-03-12
To fix issue# 1941. AMQP/S address added into discovery info response.

shwethanidd:
# Description

Discovery info response contains the AMQP/S address. That is used to build connection to remote RMQ based platform.

Fixes # (issue)
1941

### DIFF
--- a/volttron/platform/web/master_web_service.py
+++ b/volttron/platform/web/master_web_service.py
@@ -328,15 +328,13 @@ class MasterWebService(Agent):
 
         return_dict = {}
 
-        if self.serverkey:
+        if external_vip and self.serverkey:
             return_dict['serverkey'] = encode_key(self.serverkey)
-        else:
-            sk = None
+            return_dict['vip-address'] = external_vip
 
         if self.instance_name:
             return_dict['instance-name'] = self.instance_name
-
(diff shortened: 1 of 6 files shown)


volttron/platform/web/master_web_service.py
```diff
@@ -337,7 +338,15 @@ def _get_discovery(self, environ, start_response, data=None):
 
         return_dict['vip-address'] = external_vip
         if self.core.messagebus == 'rmq':
-            return_dict['vc-rmq-address'] = self.volttron_central_rmq_address
+            config = RMQConfig()
+            rmq_address = None
+            if config.is_ssl:
```
craig8: seems like you can just change the protocol in the if statement and add it as another parameter rather than the two lines that are exactly the same but the s in amqps

craig8:
@shwethanidd  if there is not a vip-address within the config file then I think the vip-address and serverkey should not show up in the payload response.  Thoughts?
image

volttron/platform/web/master_web_service.py
shwethanidd: Port also is changing with amqp/s.

craig8: Ahh ok

PR #1949, opened 2019-03-16
base_historian tests and platformwrapper improvements

schandrika:
base historian test case fixes
platformwrapper improvements - now build_agent is called only once at instance startup. Rest of the method do not create dynamic agents.


shwethanidd:
Fixes for issue#1740

PR #1952, opened 2019-03-18
Rmq tests

shwethanidd:
# Description
Fixes some of the pytest issues such as test_can_start_webserver, test_can_install_listeners, test_fixture_starts_platforms and some of the auth tests

Fixes Issue#1740.


volttrontesting/fixtures/rmq_test_setup.py
```diff
@@ -59,7 +59,10 @@ def create_rmq_volttron_setup(vhome=None, ssl_auth=False):
     # instance can get started and shutdown multiple times within test. But restore should happen only
     # at end of instance lifetime.
     conf_backup = os.path.join(os.path.dirname(vhome),"backup_rabbitmq_conf_"+ os.path.basename(vhome))
-    shutil.copy(os.path.join(rabbitmq_config["rmq-home"],'etc/rabbitmq/rabbitmq.conf'), conf_backup)
+    try:
+        shutil.copy(os.path.join(rabbitmq_config["rmq-home"],'etc/rabbitmq/rabbitmq.conf'), conf_backup)
+    except IOError as e:
+        _log.exception("rabbitmq.conf missing from path {}".format(conf_backup))
```
schandrika: should be "rabbitmq.conf missing from path {}".format(os.path.join(rabbitmq_config["rmq-home"],'etc/rabbitmq/rabbitmq.conf')).  This can only happen when rmq was run without ssl and with default ports. Are we still running tests without ssl ?

volttrontesting/fixtures/volttron_platform_fixtures.py
```diff
@@ -109,9 +110,9 @@ def cleanup():
     request.addfinalizer(cleanup)
     return wrapper
 
-
+# IPC testing is removed since it is not used from VOLTTRON 6.0
 @pytest.fixture(scope="function",
-        params=['tcp', 'ipc'])
+        params=['tcp'])
```
schandrika: If there is only one param we shouldn't really parametrize the test. Avoid the additional overload.

volttrontesting/utils/platformwrapper.py
```diff
@@ -134,9 +134,11 @@ def build_vip_address(dest_wrapper, agent):
 def start_wrapper_platform(wrapper, with_http=False, with_tcp=True,
                            volttron_central_address=None,
                            volttron_central_serverkey=None,
-                           add_local_vc_address=False):
+                           add_local_vc_address=False,
+                           instance_name='volttron1'):
```
schandrika: For tests aren't we using volttron_test as instance name? I don't want to use volttron1 as it is the default in the environment outside of pytest too.

volttrontesting/utils/platformwrapper.py
```diff
@@ -819,7 +822,10 @@ def install_agent(self, agent_wheel=None, agent_dir=None, config_file=None,
     def restore_conf(self):
         if self.rmq_conf_backup:
             _log.debug("Restoring original rabbitmq.conf to server and restarting rmq")
-            shutil.move(self.rmq_conf_backup, os.path.join(rabbitmq_config["rmq-home"],'etc/rabbitmq/rabbitmq.conf'))
+            try:
+                shutil.move(self.rmq_conf_backup, os.path.join(rabbitmq_config["rmq-home"],'etc/rabbitmq/rabbitmq.conf'))
+            except IOError as e:
```
schandrika: When does this happen - we have a valid not null self.rmq_conf_backup but the file doesn't exist?

volttrontesting/fixtures/rmq_test_setup.py
shwethanidd: I'll make the change. However this behavior is seen many times specially after running certs related tests.

volttrontesting/utils/platformwrapper.py
shwethanidd: This behavior is seen many times specially after running certs related tests.

volttrontesting/fixtures/volttron_platform_fixtures.py
shwethanidd: done

volttrontesting/utils/platformwrapper.py
shwethanidd: done

volttrontesting/utils/platformwrapper.py
shwethanidd: Removing try-except to investigate actual problem

Issue #1963, opened 2019-03-26
topic watcher agent fails test cases


schandrika:
Topic watcher fails existing test cases and lacks test cases for publishing to remote agent
1.  Topic watcher causes duplicate record in the log database
2.  When there are more than one alert group monitoring for topics does not happen every second
3. Add additional test cases for remote publish of alerts

PR #1971, opened 2019-03-28
Fix for #1963

schandrika:
Fix for #1963


craig8:
when publish-remote is false it shouldn't attempt to connect to the remote platform.

craig8:
If remote platform is not accepting connection when agent starts then it doesn't attempt to reconnect

craig8:
does not seem that it is publishing on the remote platform.

Issue #1973, opened 2019-04-01
crate historian should implement missing features 


schandrika:
crate historian should implement all abstract methods from base historian
crate historian does not pass most of the test cases in test_historian

schandrika:
1. Should handle explicit timezone aware time in query
2. Should handle custom schema name configured. Currently always uses 'historian'
3. Should handle custom table names configured.

PR #1976, opened 2019-04-08
Crate historian updates

schandrika:
Fix for #1973


craig8:
See email for test result output.

schandrika:
fixed test case and merged changes from upstream

PR #1982, opened 2019-04-15
Rabbitmq volttron test

jklarson:
Added dynamic IP and port for modbus_tk tests, formatted code
 Tested with ci-integration docker script

### DIFF
--- a/services/core/MasterDriverAgent/master_driver/interfaces/modbus_tk/tests/test_modbus_tk_driver.py
+++ b/services/core/MasterDriverAgent/master_driver/interfaces/modbus_tk/tests/test_modbus_tk_driver.py
@@ -4,15 +4,17 @@ import logging
 import time
 import json
 from random import randint
-from volttrontesting.utils.utils import get_rand_port
+from volttrontesting.utils.utils import get_rand_ip_and_port
 from volttron.platform import get_services_core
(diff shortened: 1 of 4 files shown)


craig8:
@jklarson @shwethanidd 

why didn't you use def get_rand_ip_and_port() which does a test for both port and ip if there is something already bound on it.

services/core/MasterDriverAgent/master_driver/interfaces/modbus_tk/tests/test_modbus_tk_driver.py
```diff
@@ -140,100 +147,115 @@ def agent(request, volttron_instance):
...
+        Issue a get_point RPC call for the device and return the result.
 
         @param agent: The test Agent.
         @param device_name: The driver name, by default: 'devices/device_name'.
-        @return: The dictionary mapping point names to their actual values from the RPC call.
+        @return: The dictionary mapping point names to their actual values from
+        the RPC call.
         """
-        return agent.vip.rpc.call('platform.driver', 'scrape_all', device_name).get(timeout=10)
+        return agent.vip.rpc.call('platform.driver', 'scrape_all', device_name) \
```
craig8: use known_identities https://github.com/VOLTTRON/volttron/blob/master/volttron/platform/agent/known_identities.py
rather than platform.driver

PR #1986, opened 2019-04-17
Implement MesaAgent-VUA 2.0

anhnguyen-cpi:
- Update AN2018 mesa_points.config and mesa_functions.yaml
   - Process data when MesaAgent receives a point that appears in multiple functions
   - Attach selector_block data when publishing a function with func_ref
   - Support publishing multiple curves in one function
   - Fix hanging on close_outstation
   - More regression tests and validations
   - DNP3Master: add send_read_command and send_unsolicited_response_command for future development
   - Not allowed to modify evariation and svariation
   - Update README and readthedocs


craig8:
Closing since we now have rabbit in develop

PR #1992, opened 2019-04-19
Single machine demo documentation

sgilbride:
# Description
Added a new demonstration for the simple instance of using the volttron-cfg command on a new install with one machine, including instructions for brief tests to ensure that agents started up correctly. In addition, fixed two minor syntax errors in other sections of the documentation.


docs/source/devguides/walkthroughs/SingleMachine-Walkthrough.rst
```diff
@@ -0,0 +1,162 @@
...
+
+The ``volttron-cfg`` command allows for an easy configuration of the VOLTTRON environment.
+
+.. note::
+        
+        To create a simple instance of VOLTTRON, leave the default response, or select yes (y) if prompted for a yes or no response [Y/N]. You must choose a username and password for the VOLTTRON Central admin account.
+
+A set of example responses are included here:
+
+.. code-block:: console
```
craig8: There are some missing prompts that happen in the rabbitmq-branch please update with those here.

craig8: You also might mention that these must be done in an activated environment. eg source env/bin/activate

PR #2001, opened 2019-05-03
vcfg update

sgilbride:
# Description
Adjusted workflow of vcfg.
IP, port, and certificates have been separated from do_vcp.
Only asks if this is an instance of volttron central if web functionality is enabled.


craig8:
@sgilbride at the end of vcfg we should not have a volttron running.

jhaack:
Picked rmq without having installed rabbitMQ. Got no error messages or warnings. Tried to start the platform and it hangs again without an explanation.

craig8:
@jhaack with https then the hostname should be the hostname i.e. myvm or whatever is in /etc/hosts file.  perhaps if rabbitmq and web enabled we just do that?

PR #2004, opened 2019-05-13
Darksky agent

jklarson:
# Description

Adds a new weather service agent for the Darksky weather API

## Type of change

Feature

# How Has This Been Tested?

Includes tests for the Darksky agent, as well as added and updated tests for the base weather agent.
Weather.gov agent also tested.

# Checklist:


services/core/Darksky/darksky/agent.py
```diff
@@ -0,0 +1,475 @@
...
+        if gresponse is None:
+            raise RuntimeError("get request did not return any "
+                               "response")
+        try:
+            response = json.loads(gresponse.content)
+            return response
+        except ValueError:
+            self.generate_response_error(url, gresponse.status_code)
+
+    def format_multientry_response(self, location, response, service):
```
schandrika: Why do we manipulate generation_time for forecast instead of returning just current time ? In base weather both get_remote_forecast and get_remote_current - we simply convert current time into UTC and save it to db in the right format. Why is it different for forecast of darksky ?

services/core/Darksky/darksky/agent.py
```diff
@@ -0,0 +1,475 @@
...
+        :param location: location dictionary requested by the user
+        :return: Timestamp and data for current data from the Darksky API
+        """
+        darksky_response = self.get_darksky_forecast(
+            SERVICES_MAPPING['SERVICE_CURRENT_WEATHER']['service'], location)
+        current_response = darksky_response.pop('currently')
+        # Darksky required attribution
+        current_response["attribution"] = "Powered by Dark Sky"
+        current_time = datetime.datetime.fromtimestamp(
+            current_response['time'], pytz.utc)
```
schandrika: Did you check if the time conversion happens right? I remember that that python’s datetime has issues converting naive timestamp to UTC correctly. That is why we use a different package (from tzlocal import get_localzone. Refer to test_historian or actuator agent) Is the time returned naive. Does the response specify the timezone info. If api response has timezone then you can create a date time object with the specific tzinfo and then convert to UTC.

services/core/Darksky/darksky/agent.py
```diff
@@ -0,0 +1,475 @@
...
+        if not len(service_name):
+            raise RuntimeError("{} is not a service provided by "
+                               "Darksky".format(service))
+        darksky_response = self.get_darksky_forecast(
+            SERVICES_MAPPING[service_name]['service'], location)
+        forecast_response = darksky_response.pop(SERVICES_MAPPING[service_name]
+                                                 ['json_name'])
+        forecast_data = []
+        for entry in forecast_response['data']:
+            entry_time = datetime.datetime.fromtimestamp(entry['time'],
```
schandrika: Check utc conversion

services/core/Darksky/darksky/agent.py
```diff
@@ -0,0 +1,475 @@
...
+    def query_forecast_service(self, service, location):
+        """
+        Generic method for requesting forecast data from the various RPC
+        forecast methods
+        :param service: forecast service type of weather data to return
+        :param location: location dictionary requested during the RPC call
+        :return: Timestamp and data returned by the Darksky weather API response
+        """
+        service_name = ''
+        for service_code in SERVICES_MAPPING:
```
schandrika: Why validate service name here? Isn’t generated by the code - i.e RPC call. Example - rpc call get_hourly_forecast(locations) - would call the downstream method with the location + correct service name ???

jklarson: This is a search for a mapped value rather than validation.

services/core/Darksky/darksky/agent.py
jklarson: Generation time is the time the predicted values were generated by the api -  this is api dependent, so we should use their scheme.

services/core/Darksky/darksky/data/name_mapping.csv
```diff
@@ -0,0 +1,40 @@
+Service_Point_Name,Standard_Point_Name,Service_Units,Standard_Units
+nearestStormDistance,,kilometer,
+nearestStormBearing,,,
+precipIntensity,lwe_precipitation_rate,millimeter / hour,meter / second
+precipProbability,,,
+temperature,surface_temperature,degC,degF
```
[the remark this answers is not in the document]
jklarson: There are tests for unit conversion in the base weather tests...the units name is not an included value with the points so I didn't think we could really make a test without making some unreasonable assumptions about the data.

I will update the config.

services/core/Darksky/darksky/data/name_mapping.csv
```diff
@@ -19,22 +19,22 @@ sunriseTime,,,
 sunsetTime,,,
 moonPhase,,,
 precipIntensityMax,,,
-temperatureHigh,,,
+temperatureHigh,,degC,degK
```
schandrika: This will not work with current code. The base weather test does not test this case either. Currently we assume that there would be valid values for standard point name if there is units.. we don't default to service point name if there is entry in csv with standard point name and units alone

services/core/Darksky/darksky/data/name_mapping.csv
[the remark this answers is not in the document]
schandrika: The config update will actually break code. Please add test case for this in base weather test and fix issue.

services/core/Darksky/darksky/agent.py
schandrika: Is there a reason there is SERVICES_MAPPING[
'SERVICE_DAILY_FORECAST'][
'service'] == “get_daily_forecast”
instead of using "get_daily_forecast" as key?

PR #2005, opened 2019-05-14
pubsub unsubscribe issues

### LINKED ISSUE #1591: self.vip.pubsub.unsubscribe("pubsub", None, None) creates KeyError exception when no subscriptions are present
self.vip.pubsub.unsubscribe("pubsub", None, None) when called without any present subscriptions gives a KeyError exception. 

I suggest correct behavior is to return success without an exception in this case.

jklarson:
Allowed users to remove all subscription by sending None for prefix and callback to unsubscribe, fixed rmq_pubsub test

# Description

Fixes #1591 (previously closed)

 response to https://volttron-community.slack.com/archives/C40SDK44E/p1557749359023200

## Type of change

# How Has This Been Tested?

- [X] Local tests for pubsub and rmq pubsub
- [x] Docker tests for pubsub and rmq pubsub

# Checklist:

### DIFF
--- a/volttron/platform/vip/agent/subsystems/pubsub.py
+++ b/volttron/platform/vip/agent/subsystems/pubsub.py
@@ -463,10 +463,13 @@ class PubSub(BasePubSub):
         bus_subscriptions = dict()
         if prefix is None:
             if callback is None:
-                if platform in self._my_subscriptions:
+                if len(self._my_subscriptions) and platform in \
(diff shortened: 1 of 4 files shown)


craig8:
@jklarson  
Actually this does not fix the issue completely.  If I call self.vip.pubsub.unsubscribe('pubsub', None, None) without having made any subscriptions this causes an error.

PR #2026, opened 2019-06-03
Changes to get all rmq certs tests to pass

schandrika:
Fixes in
- platform wrapper to do rmq single instance setup at init instead of platform start
- rmq config object to set defaults. Without this admin user is gets created as None
- Fixed test_platform_rmq to accommodate changes in platform wrapper, remove irrelevant test case, and added better error handling.

### DIFF
--- a/volttrontesting/platform/test_platform_rmq.py
+++ b/volttrontesting/platform/test_platform_rmq.py
@@ -68,18 +68,15 @@ def instance(request):
 
     yield instance
 
-    print("In fixture cleanup. skip clean up is {}".format(instance.skip_cleanup))
+    # reset skip_clean up based on env
+    instance.skip_cleanup = instance.env.get('SKIP_CLEANUP', False)
+    if instance.debug_mode:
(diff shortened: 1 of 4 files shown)


volttrontesting/platform/test_platform_rmq.py
```diff
@@ -328,26 +296,21 @@ def test_expired_ca_cert_after_vstart(request, instance):
...
             pytest.fail("Agent install should fail")
         except Exception as e:
-            print("Exception:", e)
-            assert True
-
-        # Can't find status. Essentially we have to create CA and reissue all client certs.
-        try:
-            instance.is_agent_running(agent)
-        except Exception as e:
+            # rabbitmq log will show a certificate expired error
```
craig8: Why aren't we using the pytest with exception wrapper rather than this?

schandrika: I catch exception explicitly so that I can check the exception message

Issue #1740, opened 2018-07-17
RabbitMQ-VOLTTRON: Existing pytests should run with ZMQ and RMQ message bus


shwethanidd:
VOLTTRON test fixtures have to be updated to run with RabbitMQ message bus. 
Identify test scripts that cannot run with RabbitMQ message bus and set them to run exclusively with ZMQ message bus.

shwethanidd:
@craig8 will be looking at using dockers to speed up test execution

PR #2027, opened 2019-06-04
Fix vc tests

craig8:
# Description
More fixes for #1740 this specifically fixes the multimessagebus tests for volttron central/volttron central platform.


volttron/platform/vip/agent/core.py
```diff
@@ -900,8 +900,10 @@ def __init__(self, owner, address=None, identity=None, context=None,
 
         # self._event_queue = gevent.queue.Queue
         self._event_queue = Queue()
+        # if address is a pika.ConnectionParameters then this is a remote connection to this
+        # rabbitmq so we let the identity be the user.
         if isinstance(self.address, pika.ConnectionParameters):
-            self.rmq_user = self.identity
+            self.rmq_user = '.'.join([instance_name, self.identity])
```
shwethanidd: This is same as the else statement below. Moreover what if instance_name is None?

craig8: Actually its not the same.  Note the difference is self.instance_name defaults above to volttron1...while this is using the instance_name that is passed into the function.  Also instance_name is only passed in when we are on a remote platform...however this logic is not what I would call foolproof so I will make a change to make this easier to track/debug.

PR #2031, opened 2019-06-05
More fixes to Issue#2011. Enables gevent version of pika to installed…

### LINKED ISSUE #2011: pika installed without --rabbitmq being specified
pika should only be installed during a bootstrap.py --rabbitmq run.  This issue also has a side effect of the platform not being able to be run without pika.

shwethanidd:
… from pypi

# Description

Gevent connection adapter version of pika (gevent-pika) is added to pypi repository. Now we can install it directly in the optional_requirements.json.

Also tested with "pip install gevent-pika"
Fixes #2011


kmonson:
Finally!

PR #2044, opened 2019-06-24
Fixed documentation for new versions of weather agent.

jklarson:
Documentation updates for WeatherDotGov and Darksky agents clarifying how to configure agents and send requests.


services/core/Darksky/README.rst
```diff
@@ -90,8 +90,8 @@ Example configuration:
         'api_calls_limit': 1000,
```
craig8: This configuration is not json.  

If you tried to load this using the following you will get an error
````
import json
with open("config") as fp:
    json.loads(fp.read())
````
Two ways to fix it.  Use yaml or change to using the output from json.dumps in this example

PR #2047, opened 2019-06-25
Added all_platform configuration to historian

schandrika:
Added all_platform configuration to historian. By default this will be false. Added test case to test this using both multiple zmq instances and multiple rmq instance. Testing with rmq, required adding federation setup capability to platform wrapper and fixing handling of os.environ when there is more than one volttron platformwrapper instance used within a test case.

### DIFF
--- a/services/core/SQLHistorian/config.sqlite
+++ b/services/core/SQLHistorian/config.sqlite
@@ -1,9 +1,9 @@
 {
-    "agentid": "sqlhistorian-sqlite",
     "connection": {
         "type": "sqlite",
         "params": {
-            "database": "~/.volttron/data/historian.sqlite"
+            "database": "/tmp/tmpGOh9f1/historian.sqlite"
         }
-    }
+    },
+    "all_platforms":"true"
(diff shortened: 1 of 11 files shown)


services/core/SQLHistorian/config.sqlite
```diff
@@ -1,9 +1,9 @@
 {
-    "agentid": "sqlhistorian-sqlite",
     "connection": {
         "type": "sqlite",
         "params": {
             "database": "~/.volttron/data/historian.sqlite"
         }
-    }
+    },
+    "all_platforms":"true"
```
craig8: This should be true rather than "true" as "false" would still evaluate to True when evaluating against python.

craig8:
````
Python 2.7.15rc1 (default, Nov 12 2018, 14:31:15) 
[GCC 7.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> bool("false")
True
````

PR #2173, opened 2019-12-10
Fix web

craig8:
Upgraded tests for web based solutions,
Added web token for auth with zmq (non-ssl) and tests for ssl with zmq web.


volttron/platform/certs.py
```diff
@@ -1022,3 +1022,182 @@ def create_root_ca(self, overwrite=True, valid_days=DEFAULT_DAYS, **kwargs):
...
+
+def _create_private_key():
+    return rsa.generate_private_key(
+        public_exponent=65537,
+        key_size=2048,
+        backend=default_backend()
+    )
+
+
+def _create_signed_certificate(ca_cert, ca_key, name,
```
schandrika: Why does repeat all the logic in create_ca_signed_certificate. Can we update parameters in create_ca_signed_certificated instead?

volttrontesting/fixtures/environment_fixtures.py
```diff
@@ -0,0 +1,117 @@
...
+
+from volttron.platform import update_platform_config
+from volttron.utils import get_random_key
+from volttrontesting.fixtures.cert_fixtures import certs_profile_1
+from volttrontesting.utils.platformwrapper import create_volttron_home
+from volttrontesting.utils.utils import get_hostname_and_random_port, get_rand_ip_and_port
+
+
+@contextlib.contextmanager
+def build_test_environment(messagebus: str, web_https=False, web_http=False, has_vip=True,
```
schandrika: Doesn't seem to be used anywhere. Where is this used?
There is another get_test_volttron_home() in web_utils which is used. Both seems similar. Can it be combined and moved to volttron_platform_fixtures?

volttrontesting/fixtures/cert_fixtures.py
```diff
@@ -0,0 +1,77 @@
+import contextlib
+from cryptography.hazmat.primitives import serialization
+import os
+from types import SimpleNamespace
+
+from volttron.platform.certs import CertWrapper
+
+
+@contextlib.contextmanager
+def certs_profile_1(certificate_dir, fqdn=None, num_server_certs=1, num_client_certs=3):
```
schandrika: Certs class can take a certs directory as init param. Can we create a certs class instance and then use certs method to create the server and client certs instead of duplicating code?

volttron/platform/certs.py
craig8: I think that I was headed to moving the create signed certificate out of the class into an _function....then the certs class would be responsible for storing it in a file.

craig8: I was actually in favor of making create_ca_signed_certificate use this rather than having it in the Certs object.  Since this will return the cert and key then we just save it in the other function.

volttrontesting/fixtures/environment_fixtures.py
craig8: This function is going to be used to build all test environments...there should not need to be any other function that both creates and sets up the volttron_home and config file for volttron home (note haven't added rabbit stuff).  It's used in some of the tests that I have on my hard drive :)...  The goal with this is to make the volttron_home things and then clean up at the end...and have it in one place that we can use not only within the platform wrapper but also from random tests.  The one in get_test_volttron_home is being replaced by this one.

volttrontesting/fixtures/cert_fixtures.py
craig8: I wanted a specific path to be able to be set...not just the identity set so that's why I did this so that I could "know" everything about where things were going.

PR #2209, opened 2020-01-31
Create requirements.py for release branch with all versions locked down

schandrika:
Fixed package names in requirements.py
Added new script to create a new requirements.py with all versions explicitly listed based on pip list


craig8:
This looks good, however the user has to install everything for everything to be replaced :(...I think this is the best way to do this however!

I think we may need a bootstrap.py --all to install everything what do you think?  #2218

PR #2219, opened 2020-02-10
initial volttron 7 raspbian instruction updates

laroque:
# Description

Adding install instruction details specific to Raspbian in volttron 7

## Type of change

documentation


laroque:
Should update to resolve #2220:
The instructions here (this PR) appear to be sufficient for that case, I'll change the instruction to say "on RPi systems" rather than "on Raspbian", but we should also be thoughtful about what distros we're officially or implicitly supporting.

PR #2246, opened 2020-03-13
Modified auth and admin page to keep track of approved ZMQ credentials

### LINKED ISSUE #2245: Accept ZMQ credential behavior not matching the accept RMQ CSR behavior on the admin web page.
On the admin pending_auth_reqs.html page, when a pending csr request is accepted, the admin page allows the credential to be denied or deleted later. When a pending zmq credential is accepted, it disappears from the admin page. RMQ CSRs and ZMQ credentials should behave in the same way.

sgilbride:
Fixes #2245


volttron/platform/auth.py
```diff
@@ -427,6 +430,14 @@ def deny_authorization_failure(self, user_id):
                     self._auth_denied.append(pending)
                     del self._auth_failures[self._auth_failures.index(pending)]
 
+        if len(self._auth_approved) > 0:
```
craig8: This if statement shouldn't be necessary

volttron/platform/auth.py
```diff
@@ -441,6 +452,13 @@ def delete_authorization_failure(self, user_id):
             for pending in self._auth_failures:
                 if user_id == pending['user_id']:
                     del self._auth_failures[self._auth_failures.index(pending)]
+
+        if len(self._auth_approved) > 0:
```
craig8: The outer if shouldn't be necessary...as you set them to a list above.

volttron/platform/web/admin_endpoints.py
```diff
@@ -184,8 +185,11 @@ def verify_and_dispatch(self, env, data):
             elif page == 'pending_auth_reqs.html':
                 self._pending_auths = self._rpc_caller.call(AUTH, 'get_authorization_failures').get()
```
craig8: Probably should make this a single call rather than 3 different ones...

volttron/platform/auth.py
sgilbride: Sounds good. They have been removed!

volttron/platform/web/admin_endpoints.py
shwethanidd: done

PR #2263, opened 2020-03-30
Vcfg zmq test

sgilbride:
Added vcfg zmq test cases.
Removed prompt for vc username and password from vcfg. 
Vcfg will now alert the user about the url for the admin page to set the username and password for VOLTTRON.


schandrika:
This is a great start.  We should keep adding more test cases as we test.

PR #2313, opened 2020-04-14
docs(README.md): Add note about deactivating environment

RichardLitt:
Doc only change. This was not obvious to me immediately.


shwethanidd:
Thank you for the edits

PR #2316, opened 2020-04-20
docs: Update listener heartbeat example

RichardLitt:
The current example does not match the heartbeat given out by the listener agent. While the time stamps are obviously not worth updating, the message format in general has changed significantly. I have added my current printouts.

Doc only fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,10 +279,10 @@ tail volttron.log
 Listener agent heartbeat publishes appear in the logs as:
 
 ```sh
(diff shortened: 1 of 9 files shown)


craig8:
I am wondering why the change from the json message to "good" has occured.  The point of the heartbeat message as it was constructed is to relay that "hey I am still here and my status is 'good' or bad or whatever"

@shwethanidd I am wondering what happened to the reporting of status changes for the heartbeat.

RichardLitt:
I can't answer that question: I don't know.

RichardLitt:
Thanks.

PR #2319, opened 2020-04-22
Fixes issue#2318 

### LINKED ISSUE #2318: Agent upgrade is failing
Agent upgrade is failing with following error. The existing agent is successfully removed but entry in auth.json is not removed. So when new agent with same VIP identity is installed, the agent shuts down as this VIP identity is already in the auth file.

File "scripts/install-agent.py", line 387, in <module>
    install_agent(opts, opts.package, opts.config)
  File "scripts/install-agent.py", line 149, in install_agent
    err_prefix="Error installing agent")
  File "/home/vuzer/volttron/volttron/platform/agent/utils.py", line 773, in execute_command
    raise RuntimeError()
RuntimeError

### Affected Version
releases/7.x

shwethanidd:
This PR fixes two underlying problems. When the agent gets upgraded, the same public key and secret key is being used. But the secret key was getting encoded twice which was causing assertion error. The second problem was "agent-data" directory was not created when agent is reinstalled to copy over saved agent data files. Fixed both the issues.

Fixes #2318

### DIFF
--- a/volttron/platform/control.py
+++ b/volttron/platform/control.py
@@ -563,6 +563,8 @@ def find_agent_data_dir(opts, agent_uuid):
         if x.endswith("agent-data"):
             agent_data_dir = os.path.join(opts.aip.agent_dir(agent_uuid), x)
             break
+    if not agent_data_dir:
+        agent_data_dir = opts.aip.create_agent_data_dir(agent_uuid)
     return agent_data_dir
 
 
--- a/volttron/platform/control.py
+++ b/volttron/platform/control.py
@@ -558,13 +558,8 @@ def restore_agent_data_from_tgz(source_file, output_dir):
 
 
 def find_agent_data_dir(opts, agent_uuid):
-    agent_data_dir = None
-    for x in os.listdir(opts.aip.agent_dir(agent_uuid)):
-        if x.endswith("agent-data"):
-            agent_data_dir = os.path.join(opts.aip.agent_dir(agent_uuid), x)
-            break
-    if not agent_data_dir:
-        agent_data_dir = opts.aip.create_agent_data_dir(agent_uuid)
+    # Find agent-data directory path, create if missing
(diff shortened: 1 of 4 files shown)


volttron/platform/control.py
```diff
@@ -563,6 +563,8 @@ def find_agent_data_dir(opts, agent_uuid):
         if x.endswith("agent-data"):
```
laroque, line 563: This condition is going to match more directory names than exactly the one we create in `aip.create_agent_data_dir`. Is that desired? If so, shouldn't `aip._get_data_dir` apply the same logic?
Either way, couldn't we skip this entire block and go directly to `create_agent_data_dir`?  In aip._get_data_dir it already checks if the dir exists and only creates if not.

volttron/platform/vip/socket.py
```diff
@@ -85,7 +85,13 @@ def nonblocking(sock):
...
 def encode_key(key):
     '''Base64-encode and return a key in a URL-safe manner.'''
-    assert len(key) in (32, 40)
+    if len(key) % 4 != 0:
+        return key
+    key = key if isinstance(key, bytes) else key.encode("utf-8")
+    try:
+        assert len(key) in (32, 40)
+    except AssertionError:
+        raise AssertionError("Assertion error while encoding key:{}, len:{}".format(key, len(key)))
```
laroque, line 94: Could you explain this logic change? the old behavior was an AssertionError if the length wasn't 32 or 40, but now we return the input with no error if the length isn't a multiple of 4. That seems quite different.

Also, wouldn't it be cleaner to replace lines 91-94 with:
`assert len(key) in (32, 40), f"Assertion error while encoding key:{key}, len:{len(key)}"`

volttron/platform/control.py
shwethanidd: done

volttron/platform/vip/socket.py
shwethanidd: done

PR #2344, opened 2020-05-18
Feature/recipes mergeable

laroque:
# Description

Adds the initial components for the deployment recipes system. It establishes the deployment directory with a standard ansible organization structure, which includes a number of available roles and three playbooks which build on those roles. This system is intended to facilitate the management of remote systems, especially when that involves multiple remote platforms. The components provided constitute fully useful playbooks, but also provides building blocks for users who may want to construct more specialized playbooks.

## Type of change

# How Has This Been Tested?

Because this system is inherently about working with remote systems, it is not clear how to test other than as an integration test. I've been working on a research cloud with multiple VMs, from my controller system I configure 5 remote platforms, each on its own VM:

* collector 1: RMQ on Ubuntu 18.04
* collector 2: ZMQ on Ubuntu 18.04
* collector 3: RMQ on CentOS 7
* collector 4: ZMQ on CentOS 7
* web: ZMQ on Ubuntu 18.04 - with web feature enabled, uses the controller's local volttron source tree

After newly provisioning the VMs, I copy my local ssh key to each (ssh-copy-id) and then run the three playbooks:
1. ensure-host-keys (optional, updates local known-hosts file for the VMs, this probably is done when you copy your ssh public keys, but is done as a test; it produces changes because it is more complete than what ssh-copy-id triggers)
2. host-config
3. install-platform

After running the above and seeing no errors reported, I ssh to each host & activate the volttron venv. I then start the platform and observe that I can use `vctl` to interact with it. There are currently no agents installed (hopefully coming soon).

# Checklist:

# Notes
This does *not* constitute the deployment recipes being fully feature-complete, but it is a foundation that should be usable as is, and which additional features can be added on top of. If we don't want to merge it into develop until the feature set is more complete, I would still like a review of this portion while the size is tractable. I'm fine with moving this PR onto a staging branch if that's desired.


craig8:
@laroque this looks like a good start!  The documentation you have written is a clear high level overview of what recipe's you have.

I would like to see a concrete scenario that walks the user through setting up a two vm use case.   Perhaps create a vagrant setup script for these or just start from having the 2 vms available with the ssh access.  Then walk the user through what commands they would run to set these machines up.

This is what I would do for the documentation at this point.

deployment/recipes/roles/set-defaults/tasks/main.yml
```diff
@@ -0,0 +1,60 @@
...
+    verbose_debug_tasks: "{{ verbose_debug_tasks | default(False) }}"
+    ## some standard ansible vars where we want to ensure assignment and defaults
+    ansible_user: "{{ ansible_user | default(ansible_user_id) }}"
+    ansible_python_interpreter: "{{ ansible_python_interpreter | default('{{discovered_interpreter_python}}') }}"
+    ## extra system-level pacakges to install
+    extra_system_packages: "{{ extra_system_packages | default([]) }}"
+    ## our custom ansible module has dependencies, install them to a venv
+    venv_for_ansible: "{{ venv_for_ansible | default('{{ ansible_env.HOME }}/ansible_venv') }}"
+    ## variables related to the version of volttron to be installed
+    volttron_use_local_source: "{{ volttron_use_local_source | default(False) }}"
```
craig8, line 18: if this is True how do you set the local source?  or will this be a path if wanting to use local source?

deployment/recipes/roles/system-dependencies/tasks/main.yml
```diff
@@ -0,0 +1,25 @@
+### A role for ensuring that all system-level dependencies are installed.
+### Many tasks in this role require admin-level access, it will make changes
+### to the system and their is no support for reverting them.
+---
+- name: initialize install tracker fackt
+  set_fact:
```
craig8, line 6: spelling fackt

deployment/recipes/roles/set-defaults/tasks/main.yml
craig8: I saw later that it uses volttron_root.  Perhaps a comment to where that's set in the host-config.

laroque: It is currently defined to be two dirs up relative to the path of the playbook. Could make it 3 options:
false -> use github
true (not a string) -> current logic, relative to playbook
true (is string) -> string is path

PR #2375, opened 2020-06-11
Bugfix/2374 pin erlang

### LINKED ISSUE #2374: Pin erlang to version that won't break ssl/rmq
On ubuntu 18.0.4 (not linux mint) there is an issue with erlang versions > 1:22.

Remove rabbitmq.conf from ~/rabbitmq_server/rabbitmq_server-3.7.7/etc/rabbitmq/ and ~/.volttron

Reproduce from the volttron root (using develop)

1. `./scripts/rabbitmq_dependencies.sh`
1. vcfg --rabbitmq single (answer prompts for rmq and fill)
1. volttron -vv

The output will give all kinds of connection errors.

After #2375 the errors will go away (using the same steps as above).

craig8:
# Description

Modified the scripts/rabbitmq_dependencies.sh script to pin the version of elang installed from erlang-solutions.com's mirror server.

This does not however fix the centos versioning.  I am not 100% sure if that specific issue is broken or not at this point.

Fixes #2374 

## Type of change

# How Has This Been Tested?

- Tested on ubuntu 18.04.3

# Requires testing on

- [x] @sgilbride Please test on your 18.04.4
- [x] @laroque Please test on debian buster (preferably both arm and x64)
- [x] @shwethanidd I believe you have centos 8/9
- [x] @jklarson Would you be willing to test on your machines any you can.


laroque:
Is there a reason why for deb-based systems we add the erlang-solutions package repo but for rpm-based we use bintray? I see rpm repo instructions (https://www.erlang-solutions.com/resources/download.html) and can test those if that's just something we haven't looked into unifying.

craig8:
I am not 100% sure.  I know that this was put at @schandrika originally and I am sure she had a reason for it (or didn't know about it possibly)?

laroque:
Did not look into the erlang-solutions vs bintray thing but did test and find bugs on RPi. I like this PR after https://github.com/craig8/volttron/pull/45 gets incorporated.

craig8:
@shwethanidd were you able to work on the centos version verification?

RichardLitt:
I can confirm that this worked for me on buster bionic.

PR #2383, opened 2020-06-19
Fixes issue#2345. 

### LINKED ISSUE #2345: Auth does not get list of peers connected to the platform during startup
# Description of Issue
During platform startup AuthService agent reads auth.json and attempts to send agent capabilities message to all the agents connected to the platform. AuthService agents tries to get list of agents using peerlist(). Peerlist() fails silently with gevent timeout error.

volttron.log shows below statement

2020-05-18 15:14:18,143 () volttron.platform.auth WARNING: Attempt 1 to get peerlist failed with exception

Related to issue #2119

### Affected Version
7.0 rc

shwethanidd:
AuthService attempts to get peerlist to send the auth updates to all it's peers. The peerlist subsystem is able to get the list of peers from router, but is unable to set the result into asyn result object properly. As a workaround, peerlist subsystem now keeps track of list of peers to make it accessible to auth service

# Description

Fixes #2345

### DIFF
--- a/volttron/platform/auth.py
+++ b/volttron/platform/auth.py
@@ -139,7 +139,7 @@ class AuthService(Agent):
                 _log.debug("Sending auth updates to peers")
                 # Give it few seconds for platform to startup or for the
                 # router to detect agent install/remove action
-                gevent.sleep(5)
+                gevent.sleep(2)
                 self._send_update()
             except BaseException as e:
                 _log.error("Exception sending auth updates to peer. {}".format(e))
--- a/volttron/platform/auth.py
+++ b/volttron/platform/auth.py
@@ -137,6 +137,9 @@ class AuthService(Agent):
         if self._is_connected:
             try:
                 _log.debug("Sending auth updates to peers")
+                # Give it few seconds for platform to startup or for the
+                # router to detect agent install/remove action
+                gevent.sleep(5)
                 self._send_update()
             except BaseException as e:
                 _log.error("Exception sending auth updates to peer. {}".format(e))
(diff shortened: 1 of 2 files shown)


volttron/platform/auth.py
```diff
@@ -137,6 +137,9 @@ def read_auth_file(self):
         if self._is_connected:
             try:
                 _log.debug("Sending auth updates to peers")
+                # Give it few seconds for platform to startup or for the
+                # router to detect agent install/remove action
+                gevent.sleep(5)
```
craig8, line 142: I think that we shouldn't do this as every time something triggers an auth read we will be sleeping for an arbitrary 5 seconds.

volttron/platform/vip/agent/subsystems/peerlist.py
```diff
@@ -146,14 +147,21 @@ def _handle_subsystem(self, message):
...
+                    self.peers_list.remove(peer)
         elif op == 'listing':
             try:
                 result = self._results.pop(message.id)
             except KeyError:
                 return
             # The response will have frames, we convert to bytes and then from bytes
             # we decode to strings for the final response.
-            result.set([arg for arg in message.args[1:]])
+            peers = [arg for arg in message.args[1:]]
```
craig8, line 162: Can you remove the comment above this as it doesn't make sense any more

volttron/platform/auth.py
shwethanidd: It is needed to detect new installed/removed agents. Perhaps I can reduce the sleep period

volttron/platform/auth.py
```diff
@@ -139,7 +139,7 @@ def read_auth_file(self):
                 _log.debug("Sending auth updates to peers")
                 # Give it few seconds for platform to startup or for the
                 # router to detect agent install/remove action
-                gevent.sleep(5)
+                gevent.sleep(2)
                 self._send_update()
```
craig8, line 143: Is there any way to detect whether this sleep is necessary or not?

shwethanidd: No. I can't of think of any way. I can remove it if it is a nuisance.

PR #2388, opened 2020-06-22
Add try/finally block to docker_wrapper

### LINKED ISSUE #2387: Make docker_wrapper cleanup containers when a test fails
NOTE issues are for bugs/feature requests not support.  Please ask at https://stackoverflow.com/questions/tagged/volttron for help

### Description of Issue
If a test uses the docker_wrapper to create a container and the test fails, the container is not killed and is left up and running. 

### Affected Version
develop branch

### Screenshots
Screen Shot 2020-06-22 at 10 18 02 AM

#### Expected
Container should be killed and removed. 

#### Actual
Container is still running.

### Steps to Reproduce
1. Create a test using docker_wrapper and make it fail
2. Check docker to see if container is still running. Use the following command:
```bash
$ docker ps -a
```

### Additional Details
In addition to cleaning up containers after a failed test, catch exceptions thrown by the Docker API and add tests.

bonicim:
# Description

Currently, if a test used docker_wrapper and the test failed, the
container will not be killed and rather will remain up and running. Thus, every time a test that uses docker_wrapper fails, the container must manually be killed/resource released by running ```docker container stop $(docker ps -aq)```. 

Adding the try/finally block will ensure that container is killed automatically.
See https://docs.python.org/3/library/contextlib.html

Fixes #2387 

## Type of change

# How Has This Been Tested?

I ran the docker wrapper tests. 

I also created a failing test, then checked to see if the container was still running. 
Then I added the fix, reran the failing test, and checked to see if the container was killed/released.

Before:
Screen Shot 2020-06-22 at 10 18 02 AM

After:
Screen Shot 2020-06-22 at 10 27 05 AM

# Checklist:

### DIFF
--- a/volttrontesting/fixtures/docker_wrapper.py
+++ b/volttrontesting/fixtures/docker_wrapper.py
@@ -36,11 +36,7 @@ if HAS_DOCKER:
             ::
                 # example port exposing mysql's known port.
                 {'3306/tcp': 3306}
-        :param env: environment variables to set inside the container, as a dictionary following the convention
-
-            ::
-                example environment variable for mysql's root password.
-                {'MYSQL_ROOT_PASSWORD': '12345'}
+        :param env: environment variables to set inside the container.
         :param command: string or list of commands to run during the startup of the container.
         :param startup_time_seconds: Allow this many seconds for the startup of the container before raising a
             runtime exception (Download of image and instantiation could take a while)
@@ -53,19 +49,13 @@ if HAS_DOCKER:
         client = docker.from_env()
 
         try:
-            repo = image_name
-            if ":" not in repo:
+            full_docker_image = image_name
+            if ":" not in full_docker_image:
                 # So all tags aren't pulled. According to docs https://docker-py.readthedocs.io/en/stable/images.html.
-                repo = repo + ":latest"
-            client.images.pull(repo)
-        except APIError as e:
-            raise RuntimeError(e)
-
-        try:
+                full_docker_image = full_docker_image + ":latest"
+            client.images.pull(full_docker_image)
             container = client.containers.run(image_name, ports=ports, environment=env, auto_remove=True, detach=True)
-        except ImageNotFound as e:
-            raise RuntimeError(e)
-        except APIError as e:
+        except (ImageNotFound, APIError, RuntimeError) as e:
             raise RuntimeError(e)
 
         if container is None:
--- a/volttrontesting/fixtures/docker_wrapper.py
+++ b/volttrontesting/fixtures/docker_wrapper.py
@@ -1,6 +1,6 @@
 try:
     import docker
-
+    from docker.errors import APIError, ImageNotFound
     HAS_DOCKER = True
 except ImportError:
     HAS_DOCKER = False
@@ -51,18 +51,39 @@ if HAS_DOCKER:
 
         # Create docker client (Uses localhost as agent connection.
         client = docker.from_env()
-        if ":" in image_name:
-            client.images.pull(image_name)
-        else:
-            # So all tags aren't pulled. According to docs https://docker-py.readthedocs.io/en/stable/images.html.
-            client.images.pull(image_name + ":latest")
-        container = client.containers.run(image_name, ports=ports, environment=env, auto_remove=True, detach=True)
+
+        try:
+            repo = image_name
+            if ":" not in repo:
+                # So all tags aren't pulled. According to docs https://docker-py.readthedocs.io/en/stable/images.html.
+                repo = repo + ":latest"
+            client.images.pull(repo)
+        except APIError as e:
+            raise RuntimeError(e)
+
+        try:
+            container = client.containers.run(image_name, ports=ports, environment=env, auto_remove=True, detach=True)
+        except ImageNotFound as e:
+            raise RuntimeError(e)
+        except APIError as e:
+            raise RuntimeError(e)
 
         if container is None:
             raise RuntimeError(f"Unable to run image {image_name}")
 
+        # wait for a certain amount of time to let container complete its build
+        try:
+            if _not_valid_container(container, startup_time_seconds):
+                yield None
+            else:
+                yield container
+        finally:
+            container.kill()
+
+    def _not_valid_container(container, startup_time_seconds):
         error_time = time.time() + startup_time_seconds
         invalid = False
+
         while container.status != 'running':
             if time.time() > error_time:
                 invalid = True
@@ -70,11 +91,4 @@ if HAS_DOCKER:
             time.sleep(0.1)
             container.reload()
 
-        try:
-            if invalid:
-                yield None
-            else:
-                yield container
-        finally:
-            container.kill()
-
+        return invalid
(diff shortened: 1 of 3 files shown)


volttrontesting/fixtures/docker_wrapper.py
```diff
@@ -31,12 +31,16 @@ def create_container(image_name: str, ports: dict = None, env: dict = None, comm
...
                 # example port exposing mysql's known port.
                 {'3306/tcp': 3306}
-        :param env:
+        :param env: environment variables to set inside the container, as a dictionary following the convention
+
+            ::
+                example environment variable for mysql's root password.
+                {'MYSQL_ROOT_PASSWORD': '12345'}
         :param command: string or list of commands to run during the startup of the container.
         :param startup_time_seconds: Allow this many seconds for the startup of the container before raising a
```
craig8, line 45: This comment makes it seem like the download is in the # seconds.  I don't believe that is the case

volttrontesting/fixtures/docker_wrapper.py
```diff
@@ -47,29 +51,44 @@ def create_container(image_name: str, ports: dict = None, env: dict = None, comm
...
-            client.images.pull(image_name)
-        else:
-            # So all tags aren't pulled. According to docs https://docker-py.readthedocs.io/en/stable/images.html.
-            client.images.pull(image_name + ":latest")
-        container = client.containers.run(image_name, ports=ports, environment=env, auto_remove=True, detach=True)
+
+        try:
+            repo = image_name
+            if ":" not in repo:
+                # So all tags aren't pulled. According to docs https://docker-py.readthedocs.io/en/stable/images.html.
```
craig8, line 58: I don't think it would pull all of the tags...we just are being explicit about pulling the latest tag.  unless somewhere on the linked page tells me otherwise....in which case you should like to the exact location in that page.

volttrontesting/fixtures/docker_wrapper.py
```diff
@@ -47,29 +51,44 @@ def create_container(image_name: str, ports: dict = None, env: dict = None, comm
...
         client = docker.from_env()
-        if ":" in image_name:
-            client.images.pull(image_name)
-        else:
-            # So all tags aren't pulled. According to docs https://docker-py.readthedocs.io/en/stable/images.html.
-            client.images.pull(image_name + ":latest")
-        container = client.containers.run(image_name, ports=ports, environment=env, auto_remove=True, detach=True)
+
+        try:
+            repo = image_name
```
craig8, line 56: consider changing repo to be full_docker_image  then comment that full_docker_image is image plus tag

but repo doesn't make sense in this context for the variable name.  At least to me

volttrontesting/fixtures/docker_wrapper.py
```diff
@@ -47,29 +51,44 @@ def create_container(image_name: str, ports: dict = None, env: dict = None, comm
...
+            if ":" not in repo:
+                # So all tags aren't pulled. According to docs https://docker-py.readthedocs.io/en/stable/images.html.
+                repo = repo + ":latest"
+            client.images.pull(repo)
+        except APIError as e:
+            raise RuntimeError(e)
+
+        try:
+            container = client.containers.run(image_name, ports=ports, environment=env, auto_remove=True, detach=True)
+        except ImageNotFound as e:
```
craig8, line 66: I think you could do 

````
except (ImageNotFound, RuntimeError) as e:
````

File: volttrontesting/testutils/test_fixtures.py [hunk 5]
```diff
@@ -1,12 +1,14 @@
 import pytest
-from volttrontesting.fixtures.volttron_platform_fixtures import volttron_instance_web
```
craig8, line 2: lets rename this file to be test_docker_wrapper instead of test_fixtures  or put the docker_wrapper tests in a new file.  I don't see all the stuff in the diff here.

volttrontesting/fixtures/docker_wrapper.py
bonicim: Agree. I was thinking of what to name this variable and looked at Python docker's images.pull(...) method for inspiration which used "repository" for its parameter. Full_docker_image seems more descriptive in this context.

volttrontesting/fixtures/docker_wrapper.py
bonicim: Yes, you're right. ImageNotFound is a NotFound, which is also an APIError. Will change.

volttrontesting/fixtures/docker_wrapper.py
bonicim: I tested out the call without ":" on the interpreter. It looks like it does pull all the images if a tag is not given. I had to kill the process since it was taking too long. Then I checked docker for images and a bunch of images appeared.
```bash
>>> import docker
>>> client = docker.from_env()
>>> image = "mysql"
>>> res = client.images.pull(image)
^CTraceback (most recent call last):

# Checking docker
$ docker images ls

REPOSITORY                                                     TAG                 IMAGE ID            CREATED             SIZE
crate                                                          latest              26e2f7f0075c        13 days ago         676MB
mysql                                                          5.7                 9cfcce23593a        2 weeks ago         448MB
mysql                                                          latest              be0dbf01a0f3        2 weeks ago         541MB
670892884287.dkr.ecr.us-gov-west-1.amazonaws.com/lessons/api   0.2.9               f24c01f59ca8        6 months ago        594MB
670892884287.dkr.ecr.us-gov-west-1.amazonaws.com/lessons/api   latest              f24c01f59ca8        6 months ago        594MB
mysql                                                          5.5.57              6929c9b0e536        2 years ago         257MB
mysql                                                          5.5.56              17e9a1fe381e        3 years ago         256MB
mysql                                                          5.5.55              c0fb485deb70        3 years ago         256MB
mysql                                                          5.5.54              9913483658d3        3 years ago         256MB
mysql                                                          5.5.53              0e21a45a5660        3 years ago         255MB
mysql                                                          5.5.52              402df65b897e        3 years ago         255MB
mysql                                                          5.5.51              51ae120c924c        3 years ago         256MB
mysql                                                          5.5.50              385c5c57528f        3 years ago         256MB
...........and more images



```

volttrontesting/testutils/test_fixtures.py
bonicim: Will do. I'll also remove the test_web_setup_properly since that doesn't test the docker_wrapper. Why was it there in the first place?

volttrontesting/fixtures/docker_wrapper.py
bonicim: Yes, it is not the case as I have learned the hard way. It could also be misleading since different images of db's have different environment variables. I'll remove the example.

bonicim:
On a different note, the parameter `command` is not used. See https://github.com/VOLTTRON/volttron/blob/develop/volttrontesting/fixtures/docker_wrapper.py#L15

What was the original reason for having it there? Perhaps as an option to pass to the 'containers.run' api call?

craig8:
Yes the command should be able to override the api call so you could pass something like

````
docker run -it ubuntu ls -la
# where ls -la would be the command
````

bonicim:
> Yes the command should be able to override the api call so you could pass something like
> 
> ```
> docker run -it ubuntu ls -la
> # where ls -la would be the command
> ```

Oh, ok. That's what I thought. To enable this capability, should docker_wrapper use that parameter when calling `container.run`? The contract of `container.run` sets command to None by default. So explicitly setting `command` to None _shouldn't_ be an issue. I could change that and even write a small test for it. That way the parameter is used and I don't get those linter warnings about unused variables.

PR #2393, opened 2020-06-26
Add level one integration tests for MySqlFuncts

### LINKED ISSUE #2391: Add level one integration tests for MySqlFuncts class
### Description of Issue
As part of the process of improving testing for Volttron, level one integration tests for all database utility classes need to be created. These tests will ensure that the underlying database interface that supports all historian agents are working as expected. Level one integration tests use Docker images of databases as stand-in production databases for testing. For this issue, add level one integration tests for MySQLFuncts, which can be [found in the volttron/platform/dbutils](https://github.com/VOLTTRON/volttron/blob/develop/volttron/platform/dbutils/mysqlfuncts.py).

### Affected Version
Develop

bonicim:
# Description

Add level one integration tests for the MySQLFuncts class which provides the interface between a MySQL database and the SQL Historian agent. 

This test uses a Docker image of MySQL to act as the production database for testing. The database is seeded with data by executing SQL statements directly in the Docker container. [MySQLConnector/Python](https://dev.mysql.com/doc/connector-python/en/) is used to verify tests by reading the database and using that output to asserting that the actual result is equal to the expected result.

This test does not test the `get_agg_topics` method because I was still having problems getting it to work but didn't want that one test to block the publication and feedback of this PR. That method will be addressed in a later PR or commit to this PR.

The tests allow the flexibility to conduct testing against a list of different versions of MySQL. Currently, there is one version in the list (5.7). The latest version, 8.0, is not tested because MySQLFuncts was designed when 5.6/5.7 was the latest version. When I ran the tests for the aggregate-related tables against the MySQL 8.0, MySQL threw out a `Specified key was too long; max key length is 3072 bytes` error. I have documented this in issue 2392.

Fixes #2391 

## Type of change

# How Has This Been Tested?

I have ran the tests on both PyCharm and the command line using Pytest. See below:

```bash
# must be in the directory of the tests
$ pytest -m mysqlfuncts

# output
======== 18 passed, 86 warnings in 205.28s (0:03:25) =====
```

# Checklist:

### DIFF
--- a/volttrontesting/fixtures/docker_wrapper.py
+++ b/volttrontesting/fixtures/docker_wrapper.py
@@ -65,11 +65,11 @@ if HAS_DOCKER:
                 break
             time.sleep(0.1)
             container.reload()
-
-        if invalid:
-            yield None
-        else:
-            yield container
-
-        container.kill()
+        try:
+            if invalid:
+                yield None
+            else:
+                yield container
+        finally:
+            container.kill()
 
--- a/volttrontesting/platform/dbutils/test_mysqlfuncts.py
+++ b/volttrontesting/platform/dbutils/test_mysqlfuncts.py
@@ -1,0 +1,606 @@
+import contextlib
+import datetime
+from time import time
+
+import pytest
+import mysql.connector
+
+from volttron.platform.dbutils.mysqlfuncts import MySqlFuncts
+from volttrontesting.fixtures.docker_wrapper import create_container
+from volttrontesting.utils.utils import get_rand_port
+
+
+# mysqlfuncts was written for MYSQL 5.7; however, the latest version is 8.0
+# these tests cannot use latest or anything 8.0 and above and will fail if the latest image/8.0 is used
+# for example, latest/8.0 will throw a "specified key was too long; max key length is 3072 bytes" error
+IMAGES = ["mysql:5.7", ]  # To test more images, add them here
+TEST_DATABASE = "test_historian"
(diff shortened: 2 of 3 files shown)


volttrontesting/fixtures/docker_wrapper.py
```diff
@@ -65,11 +65,11 @@ def create_container(image_name: str, ports: dict = None, env: dict = None, comm
...
             time.sleep(0.1)
             container.reload()
-
-        if invalid:
-            yield None
-        else:
-            yield container
-
-        container.kill()
+        try:
```
bonicim, line 68: I know that I have another PR (#2388) that adds this try/except, but I need it here so that the test can automatically cleanup containers.

volttrontesting/platform/dbutils/test_mysqlfuncts.py
```diff
@@ -0,0 +1,606 @@
...
+                   "agg_topics_table": AGG_TOPICS_TABLE,
+                   "agg_meta_table": AGG_META_TABLE}
+
+    mysqlfuncts = MySqlFuncts(connect_params, table_names)
+
+    yield mysqlfuncts
+
+
+@pytest.fixture(params=IMAGES)
+def get_container_func(request):
```
bonicim, line 451: This fixture gives the capability to test multiple versions of MySQL. It returns both the docker_wrapper function and a list of images that will be used to create a container. In other words, it acts like a for-loop for each test so that it can be run on different versions of MySQL.

PR #2404, opened 2020-07-07
Feature/recipes agents

laroque:
@craig8 This branch is usable for testing, but I haven't added any docs for the new feature and there are various `#TODO`s spread over. I think it should work for you to run on the particular case we discussed, but I don't know if we want to merge it into this preview branch yet. ...

There are some notes/questions:
1. I've followed the ansible boilerplate for documenting ansible modules. It would be great to find their sphinx plugin or whatever they use to parse that and generate docs. I haven't looked at this at all but it is on my list somewhere.
2. I previously said that the system supports installing a different volttron version on the remote than is used on the controller. That's still true, but to map vip_id to uuid I pulled in your new `--json` flag as an option in `vctl`. Configuring agents is dependent on this and parsing the old format seems like a pain to support. There are some options to deal with this:
  1. maybe the recipe should install a second vctl on the remote that it can use for this purpose?
  2. maybe it is worth writing that output parser after all?


laroque:
closing this PR, there's a new on from laroque:feature/recipes-mergeable which is more complete

PR #2405, opened 2020-07-08
Add level one integ tests for postgresqlfuncts

bonicim:
# Description

As part of the effort to improve testing of core agents in Volttron, adding level one integration tests for the Postgresfuncts class which serves as the workhorse behind the SQL Historian agent. This code is formatted according to the PEP8 compliant conventions specified in [Black](https://pypi.org/project/black/).

In addition to writing the level one integration tests, I have also modified the Dockerfile for the test_image, the `run-test-docker.sh` script, and requirements test file in order to run the test script both locally and on Travis CI.

On the Dockerfile.testing, I added two main layers that installed docker on the container in order to allow the level one integration test the ability to create Docker containers from a container. The second layer was adding the volttron user to the docker group in order to execute docker commands. 

On the test script, I updated the `run_test` function to run a certain docker run command when executing a level one integration test. I also fixed `process_pid` function to not exit cleanly if FAIL_FAST is false. Currently if FAIL_FAST is set to false, the tests still fail fast. 

Requirements were updated to take the latest pytest version and some level one integration test dependencies.  

NOTE 1: Running all the image version of Postgresql on TravisCI will cause a build timeout. At most, TravisCI can run these tests on two images. This has been verified on my  throwaway TravisCI branch: https://travis-ci.org/github/bonicim/volttron/builds/714569840. However, you can still run tests on all the images locally.

NOTE 2: `/var/run/docker.sock` has been modified in entrypoint to allow all users including Volttron user to have docker privileges. Although Volttron user was added to the Docker group, the database test container must be restarted in order fo the change to occur. However, the database test container cannot be rebooted because it results in a "System has not been booted with systemd as init system (PID 1)." This is obviously a security risk, however, given that the container that uses this `entrypoint.sh` is ephemeral and used only for testing, the risk is minimal and more importantly allows successful level one integration testing on Travis CI. There is the option to use rootless mode, https://docs.docker.com/engine/security/rootless/, but it is experimental and something to be explored later down the road.

## Type of change

# How Has This Been Tested?

These tests were ran on three different levels:

**First**, I simply ran the tests locally on the command line using Pytest CLI. Level one integration tests use a Docker container to act as a production database. In this case, Postgresqlfuncts will make calls to the postgresql database container. These level one integration tests ensure that Postgresqlfuncts works as expected.

I have ran the tests on both PyCharm and from the command line. Below is an example of running the test on the command line:

```bash
$ pytest -m postgresqlfuncts
```
Screenshot of output of tests:
Screen Shot 2020-07-07 at 5 48 09 PM

**Second**, I modified contents inside `ci-integration` to properly automate the running of this level one integration test. To run the test locally, I used the following command:

```bash
# note: must be in the top level of the working directory
# note: must set fail fast to false so we can ignore failed tests and  keep running tests to include the level one integ tests
$ export FAIL_FAST=false
$ ./ci-integration/run-test-docker.sh
``` 

**Third**, I created a "throwaway branch" on my local GH account/Travis account and modified the test script to only test for this specific level one integration test. I wanted to quickly find out if the configuration was working. To view how the test script executed in Travis, https://travis-ci.org/github/bonicim/volttron/builds/714569840.

# Checklist:


ci-integration/run-test-docker.sh
```diff
@@ -113,9 +121,11 @@ process_pid(){
                 if [[ ${FAST_FAIL} -eq 0 && -n ${CI} ]]; then
                     docker logs "${containernames[$index]}"
                 fi
-                if [[ ${FAST_FAIL} -eq 0 ]]; then
```
bonicim, line 116: The current if statement still will fail fast if the FAST_FAIL is set to false because bash treats both `true` and `false` as variables that are undefined and thus equal a length of zero. Instead, by just evaluating `FAST_FAIL` directly, we can achieve the expected behavior, i.e. when fast fail is false, don't exit cleanly. 

```bash
export FAST_FAIL=false
echo "FAST_FAIL is: $FAST_FAIL"

# Current version
if [[ ${FAST_FAIL} -eq 0 ]]; then
    echo "Current code: True. Fail fast and end the testing"
else 
    echo "Current code: False. Don't fail, keep going"
fi

# New version
if ${FAST_FAIL}; then
    echo "Updated code: True. Fail fast and end the testing"
else 
    echo "Updated code: False. Don't fail, keep going"
fi

### OUTPUT
FAST_FAIL is: false
Current code: True. Fail fast and end the testing
Updated code: False. Don't fail, keep going
```
See bash online tester link: https://rextester.com/JGGGQW47612

ci-integration/run-test-docker.sh
```diff
@@ -70,7 +70,9 @@ run_test(){
     base_filename="$(basename "$filename")"
     # Start the docker run module.
     docker run -e "IGNORE_ENV_CHECK=1" --name "$base_filename" \
-        -t volttron_test_image pytest "$filename" > "$base_filename.result.txt" 2>&1 &
+            -t --network="host" -v /var/run/docker.sock:/var/run/docker.sock volttron_test_image \
```
bonicim, line 73: Based on feedback from @craig8 from today's meeting, use same docker run command for all test runs. The additional options won't negatively affect the other tests.

volttrontesting/platform/dbutils/test_postgresqlfuncts.py
```diff
@@ -0,0 +1,880 @@
...
+# See https://volttron.readthedocs.io/en/develop/core_services/historians/SQL-Historian.html#postgresql-and-redshift
+# Since we are using TravisCI, running these tests on more than three images will stop the build because it takes more than 10 minutes to produce any log output.
+# See https://docs.travis-ci.com/user/customizing-the-build/#build-timeouts
+# Therefore, a maximum of two images are used for testing on TravisCI.
+# However, there are no limits to run this test locally with all images.
+# To run tests on all images, uncomment the commented images and run `./ci-integration/virtualization/core/entrypoint.sh'
+IMAGES = [
+    "postgres:9.6.18",
+    "postgres:10.13",
+    # "postgres:11.8",
```
craig8, line 24: Rather than doing this, can you read the environment variable TRAVIS_CI to determine if it is running on travis and then only run the first two in that case or better yet, extend the IMAGES with the others when not travis_ci.  See https://docs.travis-ci.com/user/environment-variables#default-environment-variables for default env variables.

bonicim: Using the env vars for test config makes total sense. I have made the changes and tested them on my throwaway test branch on Travis: see https://travis-ci.org/github/bonicim/volttron/builds/714970732

PR #2408, opened 2020-07-09
VCFG Tests

sgilbride:
Modified vcfg tests to include RMQ.
Added context wrapper for cleaning up temporary volttron home directories.


craig8:
Rabbitmq steps require not only the python bootstrap.py from being installed but before that the rabbit_dependencies.sh needs to be run.

PR #2415, opened 2020-07-27
Feature/recipes mergeable

laroque:
# Description

Adding recipe support for installing agents, including config store for those agents, into remote platforms.

## Type of change

# How Has This Been Tested?

Tested using vagrant-based VMs as provisioned from the examples/deployment directory

# Notes/Questions:

1. There are some ecobee-related changes in here. I don't know how they got into my branch without already being on the upstream. Are they desired changes to preserve?
2. I believe the agent install and config store are working, but expanding the documentation is still required. Opening a PR to add some visibility and for some further initial testing.
3. I restored some changes originally written by Craig to enable a `--json` flag to get json formatted output rather than the normal human-friendly table. This makes this much easier to deal with, but means that the system is unable to work with older versions of the platform that don't have this feature. Is this okay? If not I'll need to add a parser or figure out how to make the lower-level call directly.


deployment/recipes/library/volttron_agent.py
```diff
@@ -0,0 +1,633 @@
...
+        type: string
+        required: true
+    agent_spec:
+        description:
+            - a dictionary of configuration details for the agent
+            - valid keys are:
+                - agent_state: (either 'present' or 'absent'), indicates if the agent should or should not be present
+                - agent_enabled: (bool), if true, indicates that the agent should be started when the platform starts (does not imply running)
+                - agent_priority: (int; default 50) sets the priority ordering for starting an enalbed agent
+                - agent_running: (bool), if true, indicates that the agent should be started as it is installed (does not imply enabled)
```
craig8, line 108: I am not sure I like the agent_running paramenter.  Would agent_start work better here?  It seems like agent_running would be a state of the agent?

examples/deployment/recipe-inventory.yml
```diff
@@ -4,11 +4,15 @@ all:
...
       volttron_home: "~/vhome"
+      volttron_use_local_source: yes
     collector1:
       ansible_host: collector1
       volttron_use_local_source: yes
       volttron_features:
       - drivers
+      #extra_requirements:
+      #- ipython
+      #- jwt #TODO seems like this may need to be a core dependency?
```
craig8, line 15: jwt is installed in web based installations.

PR #2443, opened 2020-08-21
added workaround for duplicate data in single transaction bug

### LINKED ISSUE #2442: SQL Historian Locks up when handling duplicate data with Postgres
### Description of Issue
If data is received by the SQLHistorian with the same topic and timestamp in the same batch, inserting to the DB will fail on "CardinalityViolation: ON CONFLICT DO UPDATE command cannot affect row a second time"

### Affected Version
All

#### Expected
Data should be processed to prevent this error.

#### Actual
Historian fails to insert with below error.

### Steps to Reproduce
Disconnect Historian from DB so that it begins to cache in the local cache
Send multiple identical topic,ts pairs
Reconnect to to DB

### Additional Details
Aug 20 12:55:18 ip-10-1-2-171 python[15055]: 2020-08-20 12:55:16,199 (sqlhistorianagent-3.7.0 9516) volttron.platform.agent.base_historian ERROR: An unhandled exception occurred while publishing.
Aug 20 12:55:18 ip-10-1-2-171 python[15055]: Traceback (most recent call last):
Aug 20 12:55:18 ip-10-1-2-171 python[15055]:   File "/var/lib/volttron/volttron/platform/agent/base_historian.py", line 1113, in _do_process_loop
Aug 20 12:55:18 ip-10-1-2-171 python[15055]:     self.publish_to_historian(to_publish_list)
Aug 20 12:55:18 ip-10-1-2-171 python[15055]:   File "/var/lib/volttron/volttron/utils/docs.py", line 47, in f
Aug 20 12:55:18 ip-10-1-2-171 python[15055]:     return self.mthd(obj, *args, **kwargs)
Aug 20 12:55:18 ip-10-1-2-171 python[15055]:   File "/home/volttron/.volttron/agents/54aa5cfd-09ec-437e-9253-96bd66750e48/sqlhistorianagent-3.7.0/sqlhistorian/historian.py", line 217, in publish_to_historian
Aug 20 12:55:18 ip-10-1-2-171 python[15055]:     published += 1
Aug 20 12:55:18 ip-10-1-2-171 python[15055]:   File "/usr/lib/python2.7/contextlib.py", line 24, in __exit__
Aug 20 12:55:18 ip-10-1-2-171 python[15055]:     self.gen.next()
Aug 20 12:55:18 ip-10-1-2-171 python[15055]:   File "/var/lib/volttron/volttron/platform/dbutils/postgresqlfuncts.py", line 102, in bulk_insert
Aug 20 12:55:18 ip-10-1-2-171 python[15055]:     self.execute_stmt(query)
Aug 20 12:55:18 ip-10-1-2-171 python[15055]:   File "/var/lib/volttron/volttron/platform/dbutils/basedb.py", line 468, in execute_stmt
Aug 20 12:55:18 ip-10-1-2-171 python[15055]:     cursor.execute(stmt, args)
Aug 20 12:55:18 ip-10-1-2-171 python[15055]: CardinalityViolation: ON CONFLICT DO UPDATE command cannot affect row a second time
Aug 20 12:55:18 ip-10-1-2-171 python[15055]: HINT:  Ensure that no rows proposed for insertion within the same command have duplicate constrained values.

acedrew:
# Description
This may not be the best place to solve this problem, but I had to get a fix into production fast, and I thought I would provide it here for discussion to find the best way.

Fixes #2442 

## Type of change

# How Has This Been Tested?

- [] Test in production :) Got data flowing again.

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain:
* SDK:

# Checklist:

### DIFF
--- a/services/core/SQLHistorian/sqlhistorian/historian.py
+++ b/services/core/SQLHistorian/sqlhistorian/historian.py
@@ -178,11 +178,15 @@ class SQLHistorian(BaseHistorian):
         try:
             published = 0
             with self.bg_thread_dbutils.bulk_insert() as insert_data:
+                dedup_dict = {}
                 for x in to_publish_list:
                     ts = x['timestamp']
                     topic = x['topic']
                     value = x['value']
                     meta = x['meta']
+                    if (topic, ts) in dedup_dict:
+                        continue
+                    dedup_dict[(topic, ts)] = None
 
                     # look at the topics that are stored in the database
                     # already to see if this topic has a value


bonicim:
@acedrew I see what you mean. I think your solution makes sense especially since it follows this recommended action  on deduplication from this [explanation from pganalyze](https://pganalyze.com/docs/log-insights/app-errors/U126). I'll investigate how this affects MySQL and other db's as well.

craig8:
I believe that this will be an issue for all of the historians.  Therefore, I think it should be handled upstream of the actual concrete historians.  I don't think that the duplicates in a set of published messages should ever be able to have this happen.  Though duplicates may still happen they shouldn't in a set of published data sent to the concrete historians.

I would recommend changing the BackupDatabase class which is what the  BaseHistorian uses to manage the cache.  There are two ways I can think of to go about it, and really only one that I think fits.  

The first is when the record is written to the cache.  Making the cache "smarter" about ts -> topic_id relationships.  I don't think this is the correct place for this.  I believe that we would need more indexes on the sqlite cache and insert speed is a premium. 

The second is when the records, in batches, are retrieved.  I believe that we can do something very similar to what you did with the records in this pull request.

bonicim:
@craig8 To your second point on handling dedupes in BackupDatabase, are you suggesting to put @acedrew's solution in [this block of code](https://github.com/VOLTTRON/volttron/blob/master/volttron/platform/agent/base_historian.py#L1567-L1582)?

craig8:
Will be fixed in #2445

PR #2445, opened 2020-08-25
Filter duplicates at backup_database class

### LINKED ISSUE #2442: SQL Historian Locks up when handling duplicate data with Postgres

bonicim:
# Description

Attempts to fix data duplication error when an Historian inserts duplicate data in one transaction. In response to @craig8's [suggestion](https://github.com/VOLTTRON/volttron/pull/2443#issuecomment-679308718) of fixing the problem upstream in the BackupDatabase class, I have taken @acedrew's solution in #2443 and applied it to the backup_database class. I have refactored the method `get_outstanding_to_publish` to guarantee that the output will always be a unique list of data, (assuming the idea of "sameness" or "duplicateness" is defined as having the same topic and timestamp). Marking this PR as WIP to get feedback. 

Fixes #2442  
Related to #2443 

## Type of change

# How Has This Been Tested?

I have created a unit-ish test that initiliazes a real sqlite database with duplicate data  and then calls the `get_outstanding_to_publish` which is the method that sends data to the concrete historians. 

# Checklist:


acedrew:
This looks great, I couldn't remember if the implementation of set was O(1), but I knew dict was, that's why I used it in the heat of the moment, it looks like at this point, set has equal time performance, and far superior memory utilization to dict. 

Thanks for addressing this so quickly!

craig8:
@acedrew are you in a position to test this with your implementation, without your fix.  I am going to try it on my end and then we can move it into the develop proper.

bonicim:
I think there needs to be one more change to another method to make this complete. @craig8 mentioned it yesterday in that the backup_db needs to remove all duplicates once the publication is successful. I believe the method [`remove_successfully_published`](https://github.com/VOLTTRON/volttron/blob/master/volttron/platform/agent/base_historian.py#L1514_) needs to be modified  to delete all records that have the same topic and timestamp instead of a an `_id` field. Right now, if this is not changed, then the duplicates will remain in the cache and thus subject to being published again in the future.

bonicim:
Added one more change on backup_database in which all duplicates are removed from the cache after successful publication by historian.

volttron/platform/agent/base_historian.py
```diff
@@ -1540,19 +1540,25 @@ def remove_successfully_published(self, successful_publishes,
...
+            self._record_count -= c.rowcount
 
+        _log.debug(f"Count of rows in cache after deletion: {self._record_count}")
         self._connection.commit()
 
     def get_outstanding_to_publish(self, size_limit):
         """
-        Retrieve up to `size_limit` records from the cache.
+        Retrieve up to `size_limit` records from the cache. Guarantees a unique list of records,
+        where unique is defined as topic + timestamp.
```
craig8, line 1561: topic + timestamp is assuming concatenation...use tuple syntax (topic, timestamp) .

volttrontesting/platform/dbutils/test_backup_database.py
```diff
@@ -57,8 +59,69 @@ def test_backup_new_data_should_filter_duplicates(backup_database):
...
+        "3|2020-05-25 08:46:00|foobar_source|2|846|{}",
+    ]
+    assert outstanding_before == get_all_data("outstanding")
+
+    successful_publishes = {("2020-06-01 12:30:59", "dupetopic")}
+    expected_outstanding_after = ["3|2020-05-25 08:46:00|foobar_source|2|846|{}"]
+
+    backup_database.remove_successfully_published(successful_publishes, 1000)
+
+    assert expected_outstanding_after == get_all_data("outstanding")
```
craig8, line 99: I am wondering use id(expected_outstanding_after) != id(get_all_data("outstanding")) is true...the references in list could be the same and it shouldn't be after the serialize/deserialize from the database.

volttrontesting/platform/dbutils/test_backup_database.py
```diff
@@ -57,8 +59,69 @@ def test_backup_new_data_should_filter_duplicates(backup_database):
...
+
+
+def get_all_data(table):
+    q = f'''"SELECT * FROM {table}"'''
+    res = query_db(q)
+    return res.splitlines()
+
+
+def query_db(query):
+    output = subprocess.run(
```
craig8, line 120: I think you should be able to run this without shell = True

volttron/platform/agent/base_historian.py
```diff
@@ -1231,9 +1231,9 @@ def report_handled(self, record):
         """
         if isinstance(record, list):
             for x in record:
-                self._successful_published.add(x['_id'])
+                self._successful_published.add((x['timestamp'], x['topic']))
```
craig8, line 1234: I don't think this needed to change in the thoughts I had...I dont' think this will be good for cache as the two columns that are mentioned in the deletion is not (nor should they) be indexed.

craig8: I think that the data should be removed from the database right as a duplicate is detected rather than waiting for a record to be passed.

volttrontesting/platform/dbutils/test_backup_database.py
bonicim: I just tried checking the identity of both both object and yes, `id(expected_outstanding_after) != id(get_all_data("outstanding")) is true as expected. Then I checked the id's inside both lists and they are not the same either. Are you suggesting to check for reference on top of value in order to ensure that both objects are not the same thing?

acedrew:
> Added one more change on backup_database in which all duplicates are removed from the cache after successful publication by historian.

I'm not sure this is desired behavior, we want them to be published in order of reception, (assuming the outstanding table is ordered correctly) that way historian specific duplicate handling can happen in the historian.
For example, someone might want to on duplicate avg value or trigger an insert into a duplicate table etc. The problem is simply keeping those entries in separate transactions.

craig8:
@acedrew does my comment address your concern....I think you are right

volttrontesting/platform/dbutils/test_backup_database.py
bonicim: You're right. I just reviewed the docs on the subprocess module and realized that I should avoid `shell = True` in general.

craig8:
Duplicates always have been last one wins within a historian context... Generally duplicates are regarded as mistakes... We need to either make duplicates immutable. How did you actually run into this issue

volttron/platform/agent/base_historian.py
bonicim: I see what you mean. When I saw that the field `_successful_published` was being used to remove things from the cache, I went in the direction of removing dupes under the `remove_successfully_published` method. Will change this and remove the dupes at the point of discovery.

volttron/platform/agent/base_historian.py
```diff
@@ -1580,7 +1575,14 @@ def get_outstanding_to_publish(self, size_limit):
             headers = {} if row[5] is None else loads(row[5])
             meta = self._meta_data[(source, topic_id)].copy()
             topic = self._backup_cache[topic_id]
+
             if (topic, timestamp) in dedup:
+                _log.debug(f"Found duplicate; removing record from cache: {row}")
```
craig8, line 1580: These should be batched so that they can be deleted all through one transaction.

acedrew:
@craig8 for the SQL Historians, the enforcement of "last one wins" is the `ON CONFLICT (topic_id, ts) DO UPDATE SET value = EXCLUDED.value;` piece of the query. In theory one could replace that with `ON CONFLICT (topic_id, ts) DO UPDATE SET value = (value + EXCLUDED.value) / 2.0;` or some other update logic. If we drop the values before sending them to the DB, it would prevent these sorts of custom update logics from working. When encountering a duplicate value, it should ignore it in the current transaction, try again in a subsequent transaction, allowing the DB to manage the duplicates.
In answer to how I ran into this, I think it's because I have timestamp Naive MQTT data flowing into an instance, then being forwarded using the forward historian. The duplicates caused the issue at the top node where it was finally being sent to the SQL DB

craig8:
@acedrew we need to discuss this do you have some time tomorrow?

volttron/platform/agent/base_historian.py
bonicim: Good point. Will change.

acedrew:
@craig8 have you quit slack? Happy to follow up and discuss.

shwethanidd:
@acedrew Craig is OOO. We'll set something up when he is back tomorrow.

bonicim:
After meeting with @craig8 and @acedrew on 9/3, it was decided that duplicates should not be removed from the cache. Rather duplicates should be handled at the database level and should return duplicates to 'publish_historian' on follow-on transactions. This PR is a first attempt at doing just that. It is currently in WIP and seeking feedback on implementation and ideas for optimization/cleanup. FYI, I will be OO for the rest of the week (9/8 - 9/11).

bonicim:
Simplified tests and added some basic tests without duplicates. This PR is now ready for review.

bonicim:
After talking with @craig8 over the case of size_limit being a small number, I realized that instead of handling the duplicates after I received the data from the outstanding table, I could manipulate the SQL query to both return unique results and delete unique results by using the keyword "group by" which is similar to the "select distinct" SQL query. The updated solution eliminates the need to track duplicates and simply uses SQL to do deduplication. 

I have also added a test that does multiple transactions to show the expected behavior on a small size_limit value.

bonicim:
Added a test for sqlitehistorian unit test that tests the case in which duplicates are present in the cache. The test is still a work in progress, putting them out here for feedback.

bonicim:
As requested by @craig8 , I wrote an [sql performance script](https://gist.github.com/bonicim/9e696e287fea7bf2ebb5bbc135eda66a) to measure the difference between using 'group by' to remove duplicates using the current sql command without 'group by'. It graphs the size against the time duration. Attached is a picture of the graph. It appears that 'group by' adds about 3 seconds to the query. Two options here: 1) handle deduplication at the sql query; pros: offload work to SQL, cons: huge performance hit of about 7x; 2) handle deduplication at the Base Historian level; pros: more control, better performance maybe; cons: code bloat due to handling deduplication, increase complexity/work of Base Historian
Figure_1

bonicim:
Based on performance test of `group by` query and the feedback from the team, deduplication will now be handled at the BaseHistorian level instead of the cache database. While this improves performance, it does add more work and complexity to BaseHistorian; I have added and updated tests on both the BaseHistorian level and a concrete historian, using and sqlite-based historian.

volttron/platform/agent/base_historian.py
```diff
@@ -1572,30 +1587,43 @@ def get_outstanding_to_publish(self, size_limit):
             value = loads(row[4])
             headers = {} if row[5] is None else loads(row[5])
             meta = self._meta_data[(source, topic_id)].copy()
+            topic = self._backup_cache[topic_id]
+
+            # check for duplicates before appending row to results
+            if (topic, timestamp) in unique_records:
```
schandrika, line 1593: why not  use just topic_id, timestamp ?

volttron/platform/agent/base_historian.py
```diff
@@ -1518,41 +1520,54 @@ def remove_successfully_published(self, successful_publishes,
...
-                self._record_count -= c.rowcount
-        else:
-            temp = list(successful_publishes)
-            temp.sort()
-            c.executemany('''DELETE FROM outstanding
-                            WHERE id = ?''',
-                          ((_id,) for _id in
-                           successful_publishes))
-            self._record_count -= len(temp)
+                temp = list(successful_publishes)
```
schandrika, line 1553: there could be duplicate records even in the case were not all records in the batch got into db.

volttron/platform/agent/base_historian.py
```diff
@@ -1518,41 +1520,54 @@ def remove_successfully_published(self, successful_publishes,
...
+                if not self._dupe_ids:
+                    c.execute('''DELETE FROM outstanding
+                                WHERE ROWID IN
+                                (SELECT ROWID FROM outstanding ORDER BY ts LIMIT ?)''', (submit_size,))
+                    if self._record_count < c.rowcount:
+                        self._record_count = 0
+                    else:
+                        self._record_count -= c.rowcount
+                else:
+                    # this is a rare case, so adding logs to help track and debug when this case does occur
```
schandrika, line 1545: we could always delete based on unique_ids unless the first query would perform better. i am not sure if that is true though.

services/core/SQLHistorian/tests/test_sqlitehistorian_unit.py
```diff
@@ -0,0 +1,92 @@
...
+
+import pytest
+from gevent import sleep
+from datetime import timedelta
+from services.core.SQLHistorian.sqlhistorian import historian
+
+CACHE_NAME = "backup.sqlite"
+HISTORIAN_DB = "./data/historian.sqlite"
+
+
```
schandrika, line 13: There is a separate  test file for base historian under volttrontesting/services/historian/test_bbase_historian.py . This test case should get added to base historian tests

schandrika (changes requested):

services/core/SQLHistorian/tests/test_sqlitehistorian_unit.py
craig8: note this is a unit test not an integration test.  it does not run the platform.

bonicim: To @craig8's point, the intent of this test was to verify that a concrete a historian (in this case, an historian using sqlite) was handling duplicates as expected, i.e. that the last duplicate "wins" and gets saved in the db.

volttron/platform/agent/base_historian.py
bonicim: Good question. I originally used topic because I was using the solution from [@acedrew's earlier PR](https://github.com/VOLTTRON/volttron/pull/2443/files#diff-4d58a5e8d1e986cea80064a795a19e46d509c2a22d449c116651a121d3df514cR187), which doesn't use topic_id because the concrete historian doesn't have topic_id, only topic, which makes sense because topic_id comes from the cache, which only the base historian knows, but not a concrete historian. 

I've changed the code accordingly and all tests still pass. Topic_id is more concise to identify duplicates.

volttron/platform/agent/base_historian.py
bonicim: Can you elaborate on this particular case? How would this happen in the historian execution flow?

volttron/platform/agent/base_historian.py
bonicim: How would the query look like instead? I though I was deleting on unique_ids  using the following:

```sql
'''DELETE FROM outstanding 
  WHERE id = ?''',
((_id,) for _id in self._unique_ids)
```

schandrika: I meant you could use the above query in both if and else clause. If there are no duplicates won't unique ids simply contain all the ids the subquery (SELECT ROWID FROM outstanding ORDER BY ts LIMIT ?)''', (submit_size,)) return ?

services/core/SQLHistorian/tests/test_sqlitehistorian_unit.py
schandrika: This a good one to test. But since this feature is added to base historian it would be good to add a test case to base historian test cases irrespective of the concrete historian.

volttron/platform/agent/base_historian.py
bonicim: I see your point. I initially did it this way to highlight and isolate the case of duplicates, maintain the supposed performance gain on executing one query in the case of no duplicates, and not delete current production code. But you are right, in both cases, the query will delete items whether it has duplicates or not.

bonicim: I did a quick and dirty performance test between execute and executemany on a DELETE query with a size of 1000 records. It appears that the difference is negligible:

`executemany`
PASSED [100%]deleted
elapsed: 0.0029566287994384766

`execute`
PASSED [100%]deleted
elapsed: 0.002218961715698242

bonicim: Based on offline conversation with @schandrika, I've flattened the logic to be a simple if/else.

services/core/SQLHistorian/tests/test_sqlitehistorian_unit.py
bonicim: Agree. I added this extra test to validate our assumption that the "last one" wins in a duplicate saved to a database at the concrete historian level. The initial scope for this PR is fixing the backup_database so that duplicates aren't sent to the base_historian. I'll create a separate issue for base_historian unit tests to be tracked in a different PR.

bonicim: Added a simple base historian agent unit test that mocks the parent agent in the latest commit, 46321f8.

craig8:
@bonicim bump @schandrika status

bonicim:
@schandrika Comments added.

PR #2449, opened 2020-08-27
Fix setup of default aggregate historian tables

bonicim:
# Description

Changes setup of aggregate historian tables so that all tables are properly set. Updates the associated tests. 

Detailed explanation below:

`setup_aggregate_aggregate_historian_tables` and `read_tablenames_from_db` both rely that the metadata table have the following values for the field 'table_name': data_table, topics_table, and meta_table. If those fields are not present or correctly spelled, then the aggregate tables will not be created and the table attributes of the `postgresfuncts` instance will be set to None instead of some name. In other words, `postgresfuncts` assumes that the metadata table exists and is properly initialized. In other words, this change ensures that some default value is given instead of None when a key is missing from a dictionary; see https://docs.python.org/3/library/stdtypes.html#dict.get

## Type of change

# How Has This Been Tested?

I have updated the current level one integration tests by adding two extra cases: 1) metadata table is empty 2) metadata table does not have the correct/expected values for the the table_name field.

I was able to successfully run the tests locally using Pytest command. All other existing, untouched tests also passed.

# Checklist:

### DIFF
--- a/volttron/platform/dbutils/postgresqlfuncts.py
+++ b/volttron/platform/dbutils/postgresqlfuncts.py
@@ -164,17 +164,17 @@ class PostgreSqlFuncts(DbDriver):
             SQL('SELECT table_id, table_name FROM {}').format(
                 Identifier(meta_table_name))))
         prefix = tables.pop('', '')
-        tables['agg_topics_table'] = 'aggregate_' + tables['topics_table']
-        tables['agg_meta_table'] = 'aggregate_' + tables['meta_table']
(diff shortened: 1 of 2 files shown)


bonicim:
The Agent Development section has a subsection entitled Scripts which contains a bunch of things including Writing Agent Tests. I think a more appropriate place to put it would be in the Testing the Agent section or maybe it can have its own top level section with Agent Development. Not a big deal but comparing `develop`'s docs with this, it seems that Writing Agent Tests gets buried in Scripts. Screenshot attached. 
Screen Shot 2020-08-28 at 12 29 44 PM

PR #2460, opened 2020-09-14
Add unit marker to pytest

bonicim:
# Description

Add a 'unit' test marker in order to run all unit/level one integ tests from the command line using Pytest.

Use 'pytestmark' to mark a whole module with a custom marker. See https://docs.pytest.org/en/stable/reference.html#globalvar-pytestmark

## Type of change

# How Has This Been Tested?

Run the following on the command line; navigate to a certain directory to run all the tests under that directory

```bash
# in this example, we will run all the unit/level 1 integ tests of the Actuator agent
$ pytest -m unit services/core/AcutatorAgent

# in this example, run all the unit/level 1 integ tests of the dbutils folder
$ pytest -m unit volttrontesting/platform/dbutils 
```
# Checklist:

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -56,3 +56,4 @@ markers =
         mysqlfuncts: level one integration tests for mysqlfuncts
         postgresqlfuncts: level one integration tests for postgresqlfuncts
         dbutils: test all the level one integrations tests for dbfuncts classes
+        unit: Run all unit/level one integration tests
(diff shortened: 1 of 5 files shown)


craig8:
Do we need to add something extra to use the pytestmark global or does it come with pytest proper?

bonicim:
> Do we need to add something extra to use the pytestmark global or does it come with pytest proper?

I don't think we need to add any extra stuff to use it. From what I gather from the Pytest docs, pytest will look for a pytestmark global variable and if it's there, it will apply the marker(s) to the whole module. https://docs.pytest.org/en/stable/reference.html#globalvar-pytestmark

PR #2475, opened 2020-09-30
Updated and added VC tests.

sgilbride:
**Description**
- Updated and added VC tests. 
- This draft PR is intended to allow for others to run the VC tests in their environment to ensure that there is consistent behavior in the passing and failing tests. Inconsistent behavior is seen with cleanup between tests.

**Consistently failing with DEBUG flag and remove_agent commented out:**
1. test_vc_autoregister.py
    * test_able_to_register_unregister (sink: rmq_web; source: zmq)

2. test_webapi.py
    * test_installagent (zmq and rmq)


services/core/VolttronCentral/tests/test_vc_autoregister.py
```diff
@@ -8,8 +8,8 @@
                                                       VOLTTRON_CENTRAL_PLATFORM)
 from volttrontesting.utils.agent_additions import (add_volttron_central,
                                                    add_volttron_central_platform)
+from services.core.VolttronCentral.tests.vctestutils import APITester
```
craig8, line 11: NO this should  not have services.core in it as those should not be modules!

services/core/VolttronCentral/tests/test_webapi.py
```diff
@@ -3,17 +3,16 @@
...
     add_volttron_central_platform, add_volttron_central, add_listener
 
 from volttrontesting.utils.platformwrapper import PlatformWrapper, \
     start_wrapper_platform
 
-from zmq.utils import jsonapi
-from vctestutils import APITester
-
-from vc_fixtures import vc_and_vcp_together, vc_instance, vcp_instance
+from services.core.VolttronCentral.tests.vctestutils import APITester
```
craig8, line 13: servicces.core should not be modules!

craig8 (changes requested):

services/core/VolttronCentral/tests/test_vc_autoregister.py
craig8: The proper way to do this is to add the vctestutils to the path sys.path which would allow you to reference the vctestutils without the modules...In pycharm you could mark the tests as source to allow it on the path.

sgilbride: Reverted. Can/should the pycharm profile be added to the repo?

services/core/VolttronCentral/tests/test_webapi.py
sgilbride: Reverted.

PR #2482, opened 2020-10-07
Docs web_subsystem

sgilbride:
Updated web_subsystem documentation to describe the MANIFEST.in file, improve structure, and clarify the use of enable_web for web agent development.


docs/source/agent-framework/web-framework.rst
```diff
@@ -14,18 +14,50 @@ Web SubSystem
...
+
+.. note::
+    For all routes the first match wins.  Therefore ordering which routes are registered first becomes important.
+
 
-.. note:: The web subsystem is only available when the constructor contains enable_web=True.
+File Path
+~~~~~~~~~
+
+A path based endpoint allows the agent to specify a prefix and a static path on the file system to serve static files.
```
craig8, line 56: Note that a relative path should point to the agent-data directory within the installed agent (which is better than the example you give I think)

Also note that the user MUST have read permissions on the folder

docs/source/agent-framework/web-framework.rst
```diff
@@ -54,7 +91,11 @@ JSON endpoints allows an agent to serve data responses to specific queries from
         self.vip.web.register_endpoint(r'/vc/jsonrpc', self.jsonrpc)
 
 
-Websocket endpoints allow bi-directional communication between the client and the server.  Client connections can be authenticated during the opening of a websocket through the response of an open callback.
+Websocket
+~~~~~~~~~
+
+Websocket endpoints allow bi-directional communication between the client and the server.
+Client connections can be authenticated during the opening of a websocket through the response of an open callback.
```
craig8, line 98: In the open of the websockets, lets provide an example of how the user agent would protect against unlawful entry using the minimum requirements.

Note that there is one more aspect of this that needs to be rewritten, however I have to figure out how to serialize a Request and JsonRequest through the vip message bus.  So that agents can use those objects to return.

craig8 (changes requested):
Good job so far....just a couple of issues I have and thoughts.

docs/source/agent-framework/web-framework.rst
```diff
@@ -14,29 +14,76 @@ Web SubSystem
...
+        Allow serving of static content from 'webroot'
         """
-        self.vip.web.register_path(r'^/vc/.*', '/var/www')
+        # Sets MY_PATH to be the path of the installed agent.
+        MY_PATH = os.path.dirname(__file__)
+        # Sets WEBROOT to be the path to the webroot directory
+        # in the agent-data directory of the installed agent.
+        WEBROOT = os.path.join(MY_PATH, "webroot")
+        # Serves the static content from 'webroot' directory
+        self.vip.web.register_path("/simpleweb", WEBROOT)
```
craig8, line 79: Does this allow relative imports...eg /simpleweb/image/534.jpg?  If not please modify like the vc regular expression that was removed.

docs/source/agent-framework/web-framework.rst
```diff
@@ -54,19 +101,41 @@ JSON endpoints allows an agent to serve data responses to specific queries from
...
-    def _open_authenticate_ws_endpoint(self, fromip, endpoint):
+    def open_authenticate_ws_endpoint(self, fromip, endpoint):
         """
-        A client attempted to open an endpoint to the server.
-
-        Return True or False if the endpoint should be allowed.
-
-        :rtype: bool
+        Callback method from when websockets are opened.  The endpoine must
+        be '/' delimited with the second to last section being the session
```
craig8, line 116: seems like there is a lot of things going on in this example that aren't clear and that really don't fit.  The authentication of vc is really different than the way agents should do it.

The holding of a session is also something that should probably be broken out into an "authentication section" if there isn't one there should be

craig8 (changes requested):

craig8 (approved):

PR #2509, opened 2020-11-02
mysql create table statement fix

schandrika:
Fixed CREATE table statement for mysql historian. creating tables only if it doesn't exist. Current "CREATE TABLE IF NOT EXIST" statement require create privileges and this is not a mandatory privilege for the historian user.

Also updated corresponding documentation.


volttron/platform/dbutils/mysqlfuncts.py
```diff
@@ -151,11 +151,16 @@ def setup_historian_tables(self):
     def record_table_definitions(self, tables_def, meta_table_name):
         _log.debug(
             "In record_table_def {} {}".format(tables_def, meta_table_name))
-        self.execute_stmt(
-            'CREATE TABLE IF NOT EXISTS ' + meta_table_name +
-            ' (table_id varchar(512) PRIMARY KEY, \
-               table_name varchar(512) NOT NULL, \
-               table_prefix varchar(512));')
+
+        rows = self.select("show tables like %s", [meta_table_name])
```
craig8, line 155: If we do this check here, why are we not doing the same check for the other tables?

craig8 (changes requested):

volttron/platform/dbutils/mysqlfuncts.py
schandrika: Other tables are created in setup_historian_tables and there we check if table exists before doing create.

craig8: My question stems from the fact that you are replacing CREATE TABLE IF NOT EXIST....but you don't replace it in the create_setup_historian as well.  As you say you already know the tables don't exist...sooo?

schandrika: okay. I will change that. I was only making sure that there is a check added to see if table exists instead of relying on CREATE TABLE IF NOT EXISTS statement alone. Both record_table_definitions and setup_historian_tables now do "show tables like.." statement to check if that table exists

PR #2529, opened 2020-11-13
Fixes #2466. 

### LINKED ISSUE #2466: removing an agent with `vctl remove` fails on platform using RMQ
### Description of Issue
Using vctl to remove an agent fails with `'HTTPError'`

### Affected Version
7.x

### Steps to Reproduce
* Create a new platform using RMQ as message bus
* Install an agent (I've seen this with listener and with master driver)
* `vctl remove <uuid>` to remove the above agent
```
$ vctl remove c
Removing cee1b1da-493a-4815-a1ec-a620a3ccb29c master_driveragent-4.0
remove: error: 'HTTPError' object has no attribute 'message'
```

### Other notes
I've seen this on RPi and on an x86 VM. Using ZMQ on the same systems, removing agents works as expected.

shwethanidd:
# Description

HTTPError exception does not contain 'message' attribute. Replaced that with 'status_code'. Catching the raised exception in caller method (remove_agent)

Fixes #2466

### DIFF
--- a/volttron/platform/aip.py
+++ b/volttron/platform/aip.py
@@ -49,6 +49,7 @@ import signal
 import sys
 import uuid
 
+import requests
 import gevent
 import gevent.event
 from gevent import subprocess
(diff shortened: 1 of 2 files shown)


volttron/platform/aip.py
```diff
@@ -49,6 +49,7 @@
 import sys
 import uuid
 
+import requests
```
craig8, line 52: Shouldn't we use grequests instead of requests for this?

craig8 (approved):

PR #2535, opened 2020-12-01
Remove pubsub backward compatibility with VOLTTRON 4.1

shwethanidd:
# Description
Removed backward compatibility for pubsub over rpc feature. Pubsub through router has been in use since volttron 5 and it's time to remove support for volttron 4.1 and below versions

Fixes # JIRA_304


craig8:
Were there any specific tests wrapped around the pubsub conversion or were they assumed to work as properly if running with a 4.1 network.

PR #2537, opened 2020-12-03
Added vctl install agent-dir jira_179

craig8:
# Description

This pull request is adding vctl install agent-dir to replace the scripts/install-agent.py script.  It includes changing the tests to utilize this functionality from the command line.

## Type of change

# How Has This Been Tested?

This has been tested using the actions as a basis for the newest way of testing


shwethanidd:
Add the jia number in the description. I can point this as example to Chandrika

volttron/platform/install_agents.py
```diff
@@ -0,0 +1,361 @@
...
+            data = yaml.safe_load(fp)
+    except:
+        _log.error("Invalid yaml/json config file.")
+        sys.exit(-10)
+
+    # Configure the whl file before installing.
+    add_files_to_package(opts.package, {'config_file': config_file})
+    env = os.environ.copy()
+
+    if agent_exists:
```
shwethanidd, line 147: We should upgrade only if 'force' is provided

shwethanidd (changes requested):
Please move github workflows feature into a different PR. Is there a jira ticket for it?

volttron/platform/install_agents.py
craig8: This is actually found up above...this if is for upgrade or install.

PR #2538, opened 2020-12-04
fixes for csr create and delete certificate issues jira_306 jira_307

schandrika:
Fix for  jira_306 and jira_307

### DIFF
--- a/examples/ListenerAgent/listener/agent.py
+++ b/examples/ListenerAgent/listener/agent.py
@@ -107,7 +107,7 @@ class ListenerAgent(Agent):
         query = Query(self.core)
         _log.info('query: %r', query.query('serverkey').get())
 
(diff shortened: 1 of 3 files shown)


examples/ListenerAgent/listener/agent.py
```diff
@@ -107,7 +107,7 @@ def onstart(self, sender, **kwargs):
         query = Query(self.core)
         _log.info('query: %r', query.query('serverkey').get())
 
-    @PubSub.subscribe('pubsub', '')
```
craig8, line 110: Why changing behavior

schandrika: This is an example agent so it is good to include the all_platforms feature also in it. An example code to subscribe to all platforms..

PR #2550, opened 2020-12-28
Pin Pyjwt version

schandrika:
Fix for Jira issue 335


craig8:
@schandrika  does this require a pin to cryptography as well?

schandrika:
@craig8 We are already pin cryptography to version 2.3

PR #2551, opened 2020-12-29
RTD pages for api doc and README for volttron/services/core Agents

schandrika:
# Description

1. Converted README.rst to README.md for agents in volttron/services/core 
2. All services/core agents now have a sample config file named config and a README.md
2.  Included api documentation and README pages in read the docs for volttron/services/core agents
3. Changed md parser for sphinx
4. Pinned sphinx version in docs to force RTD to use the current sphinx version
5. Updated sphinx mocked packages


craig8:
The requirements.txt in the docs folder isn't the requirements that is loaded when running python3 bootstrap.py --documentation so the m2r2 and sphinx stuff added doesn't work as expected.  

The current --documentation should go through and install everything in the requirements.py file so if you moved the new requirements there we can remove the requirements.txt file from the docs folder.  I would suggest that.

PR #2559, opened 2021-01-12
Updated Dynamic RPC Authorizations

sgilbride:
- Updated auth entries to include rpc_method_authorizations. This is a dictionary that contains all rpc exports for an agent as the key and the authorizing capabilities for those methods in a list as the value.

- Updated auth entries to include identity. The identity is the agent_id. This value is not required, unlike the user_id field, which is usually the agent_id, but does not need to be,

- Added set and get rpc authorization methods for the auth subsystem. These can be called with 'get_rpc_authorizations" and 'set_rpc_authorizations"

- AuthService is now starts up before ConfigStoreService. This lets all other agents make rpc calls to AuthService to sync rpc method authorizations between agents and the auth file.

- VCTL can be used to dynamically modify rpc method authorizations. VCTL modifies the auth file to invoke this change. The format is vctl auth rpc allow agent_id.method rpc_method_auth1 rpc_method_auth2 ...

- Whenever the auth file is modified, (using vctl or manually) a flag is set. The rpc method authorizations in the auth file are compared to those of the active agents with the same id. If there are any changes, then the auth service updates the respective agent based on the auth file entry.

- Added pytests to test new rpc methods and other new auth features.


volttron/platform/control.py
```diff
@@ -856,6 +862,42 @@ def update_health_cache(opts):
...
+                elif not entry.rpc_method_authorizations[agent_method]:
+                    entry.rpc_method_authorizations[agent_method] = added_auths
+                else:
+                    entry.rpc_method_authorizations[agent_method].extend(
+                        [rpc_auth for rpc_auth in added_auths
+                         if rpc_auth in added_auths and
+                         rpc_auth not in entry.rpc_method_authorizations[agent_method]])
+                auth_file.update_by_index(entry, entries.index(entry))
+                return
+        _log.error(f"Agent identity not found in auth file!")
```
bonicim, line 893: Not sure if I'm reading this right, but this log error will be written every time this function is run. Is that the expected behavior?

sgilbride: Good question, that is correct! If you take a look at the for loop before this error log entry, if the agent's identity is found within the auth file during the for loop, then the function will return, leaving the function before the log statement is written. However, if the agent's identity is not found in the auth file, then the for loop will run it's course without doing anything, and the error statement will be written to the log file.

bonicim: Ah, yes, you're right. I was reading too fast and didn't see `return`. FWIW, looking at the code more closely, would moving the log error to an if-statement that finds the targeted agent identity be more expressive?  For example, before entering the for loop, a check can be made up front (similar to a guard clause) such as the following (note: I might be iterating through the list of entries incorrectly, but you get the idea):

```python
entries = auth_file.read_allow_entries()

if agent_id not in set(e.identity for e in entries):
    _log.error(f"Agent identity not found in auth file!")
    return

print(f'agent_id: {agent_id}\nEntries are:')
for entry in entries:
    ....
return
```

PR #2571, opened 2021-01-20
VCTL Auth Publickey Fix

sgilbride:
Modified filter_agents to filter based on either uuid or agent_name.
If uuid does not find any matches, then agent_name will be compared.

This change only applies to cases where no options are provided, i.e. '--tag'


volttron/platform/control.py
```diff
@@ -531,11 +531,18 @@ def filter_agents(agents, patterns, opts):
     for pattern in patterns:
         regex, _ = escape(pattern)
         result = set()
+
+        # if no option is selected, try matching based on uuid
         if not (by_uuid or by_name or by_tag):
```
craig8, line 536: what about identity? or is that a larger task

shwethanidd (approved):

PR #2573, opened 2021-01-22
Support yml json config

craig8:
# Description

This PR allows a default config, config.yml or config.json file to be loaded when running pycharm-launch script.  

## Type of change

-

### DIFF
--- a/scripts/pycharm-launch.py
+++ b/scripts/pycharm-launch.py
@@ -73,11 +73,17 @@ Optional Environmental Variables
 sys.path.insert(0, abspath)
 if not parsed.no_config:
     if not os.environ.get('AGENT_CONFIG'):
-        if not os.path.exists(os.path.join(abspath, 'config')):
+        path_found = None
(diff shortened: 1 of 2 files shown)


scripts/pycharm-launch.py
```diff
@@ -73,11 +73,17 @@ def write_required_statement(out=sys.stderr):
 sys.path.insert(0, abspath)
 if not parsed.no_config:
     if not os.environ.get('AGENT_CONFIG'):
-        if not os.path.exists(os.path.join(abspath, 'config')):
+        path_found = None
+        # Order of search is as follows config, config.yml, config.json
```
shwethanidd, line 77: It would be worth mentioning this in readthedocs

shwethanidd (commented):
Can you add this statement in readthedocs for the users to know the order of search.

scripts/pycharm-launch.py
craig8: yup I'll add it

PR #2580, opened 2021-01-26
Forwarder rtd

schandrika:
Updated forward historian documentation
Deleted unused init/config parameter (destination_instance_name) from forward historian


docs/source/agent-framework/historian-agents/forwarder/forward-historian.rst
```diff
@@ -4,58 +4,273 @@
...
 .. code-block:: json
 
     {
-        "agentid": "forwarder",
-        "destination-vip": "tcp://127.0.0.1:22916",
-        "destination-serverkey": "<SOME_KEY>"
+        "destination-address": "https://centvolttron2:8443"
     }
 
+An example configuration with above parameters is available at  `services/core/ForwardHistorian/config_web_address`.
```
shwethanidd, line 96: I don't see this file.

docs/source/agent-framework/historian-agents/forwarder/forward-historian.rst
```diff
@@ -4,58 +4,273 @@
...
 
-Adding the configuration option below will limit the backup cache to `n` gigabytes.  This will help keep a hard drive
-from filling up if the agent is disconnected from its target for a long time.
+Optional Configurations
+-----------------------
 
-::
+The most common use case for a forwarder is to forward data collected on a local instance to a remote historian. Due to
```
shwethanidd, line 101: Is there any other usecase?

schandrika: In theory I can think of other cases where the target is not a historian -- Forward analysis result to a different machine for post processing with a software that is not on the collector box? Forward simulation result to another machine that is connected to a device that can be sent a set_point instruction based on simulation result ?

docs/source/agent-framework/historian-agents/forwarder/forward-historian.rst
schandrika: Added the example config file

docs/source/agent-framework/historian-agents/forwarder/forward-historian.rst
shwethanidd: Analysis/simulation result is also data collected right? I didn't quite understand by - "use case for a forwarder is to forward data collected on a local instance". Anyway will pull it in.

PR #2584, opened 2021-01-27
Fix vctl to handle systemexit from sub commands properly

craig8:
# Description

Allows subprocessor functions to call sys.exit(0) and have that short circuit and exit properly.


sgilbride:
I have tested this with RMQ in my environment.
 - vctl install /examples/ListenerAgent/ works
 - vcfg works with all agents.

Issue #2444, opened 2020-08-24
Dates in historian out of date


craig8:
Historians should protect themselves by checking the current system date and the date of a sensor input.

Raising an alert if the system is out of date.

PR #2598, opened 2021-02-07
fix for task VOLTTRON#2444

schandrika:
Fix for issue #2444


docs/source/agent-framework/historian-agents/historian-framework.rst
```diff
@@ -67,6 +67,9 @@ All Historians support the following settings:
         # Defaults to no limit.
         "backup_storage_limit_gb": 8.0,
 
+        # How full should the backup storage be for an alert to be raised
```
craig8, line 70: is 0.9 a percentage of fullness?  What does the 0.9 mean?

volttron/platform/agent/base_historian.py
```diff
@@ -416,6 +419,10 @@ def __init__(self,
             STATUS_KEY_CACHE_FULL: False
         }
         self._all_platforms = bool(all_platforms)
+        self._time_tolerance = float(time_tolerance) if time_tolerance else None
```
craig8, line 422: what should happen if invalid input...if float time_tolerance raises an error...I think the agent should die.

volttron/platform/agent/base_historian.py
```diff
@@ -1478,39 +1544,66 @@ def free_count():
...
                     if f >= min_free_pages:
                         break
                     _log.debug(f" Cleaning cache since we are over the limit. "
                                f"After delete of 100 records from cache"
-                               f" record count is {self._record_count} page count is {p} freelist count is{f}")
+                               f" record count is {self._record_count} time_error record count is {error_record_count} "
+                               f"page count is {p} freelist count is{f}")
 
             except Exception as e:
-                _log.warning(f"Exception when check page count and deleting{e}")
```
craig8, line 1507: lets use _log.exception("Unhandled exception happened")  

No as e is necessary and it will give us a nice stack trace.

per https://www.loggly.com/blog/exceptional-logging-of-exceptions-in-python/

craig8 (changes requested):

docs/source/agent-framework/historian-agents/historian-framework.rst
schandrika: It is the decimal representation of percentage full. Updated documentation

volttron/platform/agent/base_historian.py
schandrika: If there is  invalid values for any parameter base historian init will fail and hence agent will not start. 
Updated init to also check type of time_tolerance_topics parameter. Updated configure method to also do similar check. If bad value is sent for config update using config store then, config update logs exception and fails but agent continues to run with the previous configurations

volttron/platform/agent/base_historian.py
schandrika: fixed

PR #2622, opened 2021-02-24
Pytest fix

shwethanidd:
# Description

Pytest fixes for simple failover and topic watcher tests.

### DIFF
--- a/services/ops/TopicWatcher/tests/test_topic_watcher.py
+++ b/services/ops/TopicWatcher/tests/test_topic_watcher.py
@@ -301,8 +301,6 @@ def test_watch_topic_new_group(volttron_instance, agent, cleanup_db):
 
     assert "Topic(s) not published within time limit: ['fakedevice', " \
            "'fakedevice2/all', ('fakedevice2/all', 'point')]" in alert_messages or \
-           "Topic(s) not published within time limit: ['fakedevice', " \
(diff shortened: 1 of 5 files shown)


services/ops/TopicWatcher/tests/test_topic_watcher.py
```diff
@@ -298,10 +298,14 @@ def test_watch_topic_new_group(volttron_instance, agent, cleanup_db):
     gevent.sleep(6)
 
     assert len(alert_messages) == 2
+
     assert "Topic(s) not published within time limit: ['fakedevice', " \
            "'fakedevice2/all', ('fakedevice2/all', 'point')]" in alert_messages or \
            "Topic(s) not published within time limit: ['fakedevice', " \
-           "('fakedevice2/all', 'point')], 'fakedevice2/all'" in alert_messages
+           "('fakedevice2/all', 'point')], 'fakedevice2/all'" in alert_messages or \
+           "Topic(s) not published within time limit: ['fakedevice', ('fakedevice2/all', 'point'), " \
```
schandrika, line 306: The check you added is the right one. the one above does not have the list right.  this test case only has different order of 
[ 'fakedevice', 'fakedevice2/all' , (fakedevice2/all, point) ] because order is not fixed. But it should be one list.

PR #2628, opened 2021-02-26
VC and Web SSL test fixes

sgilbride:
Modifications/Fixes:
- Updated zmq for multi-messagebus fixture to use ssl.
- Removed extraneous uses of create_web_admin.
- Refactored how web admin is set in platformwrapper. Now only sets once.
- Reworked test_admin to include two tests, test_can_authenticate_admin_user and test_can_create_admin_user.
- test_can_create_admin_user is set to skip at this time (Needs to be reworked since create_web_admin is already called in platformwrapper).
- Added verify to requests.post in vctestustils to resolve openssl issue.
- Modified cleanup_wrapper to ensure platform_wrapper is terminated (may help with cleanup issues)
- Modified VolttronCentral jsonrpc unhandled exception case to properly propagate the error message through.
- Added certs_profile_2 for returning a test certificate bundle when a contextmanager is not applicable.

Known issue:
 In test_vc_autoregister.py, the final test (zmq_web sink, rmq source) will fail regularly. However, when the specific test instance is ran independently it will pass. This is likely a cleanup issue.

### DIFF
--- a/volttrontesting/fixtures/cert_fixtures.py
+++ b/volttrontesting/fixtures/cert_fixtures.py
@@ -4,7 +4,10 @@ import os
 from types import SimpleNamespace
 
 from volttron.platform.certs import CertWrapper
-from volttron.platform.certs import Certs
+from volttron.platform.certs import Certs, _load_key
+
+#TODO: Combine cert_profile_1 and cert_profile_2
+# Verify whether we need it as dictionary or SimpleNamespace
 
 
 @contextlib.contextmanager
@@ -51,6 +54,7 @@ def certs_profile_1(certificate_dir, fqdn=None, num_server_certs=1, num_client_c
 
     yield ns
 
+
 def certs_profile_2(certificate_dir, fqdn=None, num_server_certs=1, num_client_certs=3):
     """
     Profile 2 generates the specified number of server and client certificates
@@ -72,8 +76,15 @@ def certs_profile_2(certificate_dir, fqdn=None, num_server_certs=1, num_client_c
             'O': 'pnnl',
             'OU': 'volttron_test',
             'CN': "myca"}
-    ca_cert, ca_pk = certs.create_root_ca(**data)
-
+    if not certs.ca_exists():
+        ca_cert, ca_pk = certs.create_root_ca(**data)
+    # If the root ca already exists, get ca_cert and ca_pk from current root ca
+    else:
+        ca_cert = certs.cert(certs.root_ca_name)
+        ca_pk = _load_key(certs.private_key_file(certs.root_ca_name))
+    # print(f"ca_cert: {ca_cert}")
+    # print(f"ca_pk: {ca_pk}")
+    # print(f"ca_pk_bytes: {certs.get_pk_bytes(certs.root_ca_name)}")
     ns = dict(ca_cert=ca_cert, ca_key=ca_pk, ca_cert_file=certs.cert_file(certs.root_ca_name),
                          ca_key_file=certs.private_key_file(certs.root_ca_name), server_certs=[], client_certs=[])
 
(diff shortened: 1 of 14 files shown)


volttrontesting/utils/platformwrapper.py
```diff
@@ -861,6 +874,11 @@ def startup_platform(self, vip_address, auth_dict=None,
                 # self.dynamic_agent.vip.pubsub.subscribe('pubsub', '', subscribe_to_all).get()
 
             if bind_web_address:
+                # Now that we know we have web and we are using ssl then we
```
schandrika, line 877: shouldn't this be part of init and not start if we want to create the admin user just once. There are testcases that restart platform wrapper within a testcase

volttrontesting/utils/platformwrapper.py
```diff
@@ -1507,25 +1521,23 @@ def create_web_admin(self, username, password, messagebus='rmq'):
...
+
+        # params = urlencode(dict(username='admin', password1='admin', password2='admin'))
+        # env = get_test_web_env("/admin/setpassword", method='POST')  # , input_data=input)
+        # adminep = AdminEndpoints()
+        # resp = adminep.admin(env, params)
+        # # else:
+        data = dict(username=username, password1=password, password2=password)
+        url = self.bind_web_address + "/admin/setpassword"
+        # resp = requests.post(url, data=data,
+        # verify=self.certsobj.remote_cert_bundle_file())
```
schandrika, line 1535: how does it work in case of zmq with http

schandrika (changes requested):

volttrontesting/fixtures/cert_fixtures.py
```diff
@@ -50,3 +50,46 @@ def certs_profile_1(certificate_dir, fqdn=None, num_server_certs=1, num_client_c
         ns.client_certs.append(cert_ns)
 
     yield ns
+
```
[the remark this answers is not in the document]
sgilbride: cert_profile_2 allows for use without a context wrapper. cert_profile_2 also checks if root ca for vhome already exists. Added note to combine cert_profile_1 and cert_profile_2 if possible.

volttrontesting/fixtures/volttron_platform_fixtures.py
```diff
@@ -51,7 +51,7 @@ def cleanup_wrapper(wrapper):
     # Shutdown handles case where the platform hasn't started.
     wrapper.shutdown_platform()
     if wrapper.p_process is not None:
-        if psutil.pid_exists(wrapper.p_process.pid):
+        while psutil.pid_exists(wrapper.p_process.pid):
```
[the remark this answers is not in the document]
sgilbride: It looks like platform wrapper already has that functionality. Reverted this change.

volttrontesting/utils/platformwrapper.py
sgilbride: This needs to be handled during the instance. However, cert_profile_2 now checks if root ca is already created to avoid this issue. create_web_admin handles the case of repeated requests to create a new web admin within the same instance already.

volttrontesting/utils/platformwrapper.py
sgilbride: Modified to check if ssl enabled or not. Will not use root ca to verify request if ssl is not enabled.

volttrontesting/fixtures/volttron_platform_fixtures.py
```diff
@@ -240,7 +240,7 @@ def volttron_instance_rmq(request):
 
 @pytest.fixture(scope="module",
                 params=[
-                    dict(messagebus='zmq', ssl_auth=False),
+                    dict(messagebus='zmq', ssl_auth=True),
```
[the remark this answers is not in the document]
sgilbride: HTTPS is functional in the platformwrapper for both rmq and zmq cases. Adding in additional test cases will be added during the restructuring of the tests.

shwethanidd (approved):

Issue #2472, opened 2020-09-27
Python 3.8 compatibility issues.


tirkarthi:
### Description of Issue

Python 3.8 introduced below changes that could be incompatible with current code.

* Importing ABC directly from collections has been deprecated and will be removed in Python 3.10
* `Threading.Thread.isAlive` has been deprecated and removed in favour of alias `Threading.Thread.is_alive`.
* `base64.encodestring` has been deprecated and removed in favor of `base64.encodebytes`

Following lines of code might be incompatible.

### Steps to Reproduce

```
rg -t py 'time.clock|encodestring|decodestring|celementtree|collections.*(Awaitable|Coroutine|AsyncIterable|AsyncIterator|AsyncGenerator|Hashable|Iterable|Iterator|Generator|Reversible|Sized|Container|Callable|Set|MutableSet|Mapping|MutableMapping|MappingView|KeysView|ItemsView|ValuesView|Sequence|MutableSequence|ByteString)|asyncio.?coroutine|isAlive|readfp'
volttron/platform/main.py
898:            if not thread.isAlive():
927:            if not thread.isAlive():

services/core/MasterDriverAgent/master_driver/interfaces/modbus_tk/server.py
195:        return self._server._thread.isAlive()

services/core/MasterDriverAgent/master_driver/interfaces/modbus_tk/maps/__init__.py
41:from collections import Mapping

services/core/DNP3Agent/dnp3/mesa/functions.py
62:class FunctionDefinitions(collections.Mapping):

services/core/VolttronCentralPlatform/vcplatform/agent.py
1161:                        base64.decodestring(
```

bonicim:
I checked the develop and main branch for the other two issues:
* Importing ABC directly from collections has been deprecated and will be removed in Python 3.10
* base64.encodestring has been deprecated and removed in favor of base64.encodebytes

The first issue, importing ABC directly, is still present, but 'base64.encodestring' is no longer used, making that issue moot.

PR #2664, opened 2021-04-24
Fixed deprecated function call to threading library. The old name no …

davidraker:
…longer works as of Python 3.9.

# Description

VOLTTRON does not start in Python 3.9 due to removal of a deprecated function name from the threading library used in main.py (and in modbus_tk/server.py, though that doesn't affect startup).  This PR updates the function name to one which should be compatible for all Python 3 versions.

The deprecated function is threading.Thread.isAlive(). The new name of the function has been changed to threading.Thread.is_alive(). Both versions worked prior to Python 3.9, but the deprecated version has now been removed. The function name has been updated in three places: two in main.py, and one in modbus_tk/server.py.

Fixes # (issue)
#2472 (There is more in that ticket, which is not addressed here).
## Type of change
Deprecation fix.

# How Has This Been Tested?

 - VOLTTRON starts with the fix on Fedora 33 (Python version 3.9.2), but not without.

# Checklist:

### DIFF
--- a/services/core/PlatformDriverAgent/platform_driver/interfaces/modbus_tk/server.py
+++ b/services/core/PlatformDriverAgent/platform_driver/interfaces/modbus_tk/server.py
@@ -192,7 +192,7 @@ class Server (object):
         self._server.stop()
 
     def is_alive(self):
-        return self._server._thread.isAlive()
+        return self._server._thread.is_alive()
 
     def set_verbose(self, on=True):
         self._server.set_verbose(on)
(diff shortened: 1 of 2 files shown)


craig8:
Note this works in 3.6 (https://docs.python.org/3.6/library/threading.html)

craig8:
@davidraker can you write the description of the pr to be more in line with what the pr does?

davidraker:
Craig, I added some more detail to the description, is that better?

PR #2666, opened 2021-04-29
Dynamic RPC Authorizations

sgilbride:
- Updated auth entries to include rpc_method_authorizations. This is a dictionary that contains all rpc exports for an agent as the key and the authorizing capabilities for those methods in a list as the value.
- Updated auth entries to include identity. The identity is the agent_id. This value is not required, unlike the user_id field, which is usually the agent_id, but does not need to be,
- Added set and get rpc authorization methods for the auth subsystem. These can be called with "get_rpc_authorizations" and "set_rpc_authorizations"
- VCTL can be used to dynamically modify rpc method authorizations. VCTL calls auth service to modify the auth file to evoke this change. The format is vctl auth rpc allow agent_id.method rpc_method_auth1 rpc_method_auth2 ...
- Whenever the auth file is modified, the rpc method authorizations in the auth file are compared to those of the active agents with the same id. If there are any changes, then the auth service updates the respective agent based on the auth file entry.


volttron/platform/auth.py
```diff
@@ -131,6 +147,196 @@ def setup_zap(self, sender, **kwargs):
...
+                    # Check if the rpc method does not have any rpc capabilities
+                    if not entry.rpc_method_authorizations[method]:
+                        # Set it to have the provided rpc capabilities
+                        entry.rpc_method_authorizations[method] = rpc_methods[method]
+                    # Check if the rpc method's capabilities match what have been provided
+                    if entry.rpc_method_authorizations[method] != rpc_methods[method]:
+                        # Update rpc_methods based on auth entries
+                        updated_rpc_methods[method] = entry.rpc_method_authorizations[method]
+                # Update auth file and return rpc_methods
+                self.auth_file.update_by_index(entry, entries.index(entry))
```
shwethanidd, line 208: Entry to the auth file should be made only if there is any change

volttron/platform/auth.py
```diff
@@ -131,6 +147,196 @@ def setup_zap(self, sender, **kwargs):
...
             self.vip.peerlist.onadd.connect(self._check_topic_rules)
 
+    def get_entry_rpc_method_authorizations(self, peer_id):
+        rpc_method_authorizations = {}
+        peer_methods = []
+        try:
+            peer_methods = self.vip.rpc.call(
+                peer_id, 'inspect').get(timeout=4)["methods"]
+        except gevent.Timeout:
+            _log.warn(f'{peer_id} has timed out while attempting to get rpc methods')
```
shwethanidd, line 157: change warn to warning

volttron/platform/auth.py
```diff
@@ -131,6 +147,196 @@ def setup_zap(self, sender, **kwargs):
...
+        except Unreachable:
+            _log.warn(f'{peer_id} is unreachable while attempting to get rpc methods')
+        except MethodNotFound as e:
+            _log.error(e)
+
+        for method in peer_methods:
+            try:
+                rpc_method_authorizations[method] = \
+                    self.vip.rpc.call(peer_id, 'get_rpc_authorizations', method).get(timeout=4)
```
shwethanidd, line 166: Why do we need to make rpc call to agent for every method in the inspect list? This is a costly operation since it needs to be done for every agent in the auth list.

volttron/platform/auth.py
```diff
@@ -131,6 +147,196 @@ def setup_zap(self, sender, **kwargs):
...
+                        if rpc_method_authorizations[method] is not None:
+                            # Set it to have the rpc capabilities provided by the agent if it is available
+                            entry.rpc_method_authorizations[method] = rpc_method_authorizations[method]
+                        else:
+                            entry.rpc_method_authorizations[method] = []
+                    # Check if the rpc method's capabilities match what have been provided and the agent was reachable
+                    if entry.rpc_method_authorizations[method] != rpc_method_authorizations[method] and rpc_method_authorizations[method] is not None:
+                        try:
+                            self.vip.rpc.call(
+                                entry.identity, "set_rpc_authorizations",
```
shwethanidd, line 276: Can we collect all the methods that need new authorizations and sent it to agent and let subsystem handle it?

volttron/platform/control.py
```diff
@@ -746,6 +748,10 @@ def list_agents_rpc(opts):
             try:
                 peer_method_metadata[peer][method] = conn.server.vip.rpc.call(
                     peer, f'{method}.inspect').get(timeout=4)
+                authorized_capabilities = conn.server.vip.rpc.call(
```
shwethanidd, line 751: Instead of 2 rpc calls to agent, can we have 1 rpc call in agents auth subsystem that does both the job?

shwethanidd (changes requested):

volttron/platform/auth.py
sgilbride: Fixed

volttron/platform/auth.py
sgilbride: Fixed

volttron/platform/auth.py
sgilbride: Fixed

volttron/platform/auth.py
sgilbride: Fixed.

volttron/platform/control.py
sgilbride: It is possible, though with how inspect is used, and since this is a command line check usually only calling one or two methods, I don't think it causes a large drain on resources?

shwethanidd (approved):

docs/source/platform-features/control/authentication-commands.rst
```diff
@@ -140,24 +148,31 @@ It can also take a regular expression representing a range of IP addresses.
...
+An identity is a string that represents the agent's VIP identity. This is an optional field, used by the platform to
+communicate between agents via RPC calls. It should be set if an agent has RPC exports.
+
+
 Capabilities:
 -------------
-Capability is an arbitrary string used by an agent to describe its exported RPC method. It is used to limit the access
-to that RPC method to only those agents who have that capailbity listed in their authentication record.
-
+A capability is an arbitrary string used by an agent to constrain its exported RPC method.
```
craig8, line 164: I wouldn't say its arbitrary since it still has rules that it must follow.

docs/source/platform-features/control/authentication-commands.rst
```diff
@@ -462,3 +476,37 @@ A request can be deleted using the following command:
...
+        "RPC_exported_method_2": [
+            "authorized_capability_3",
+            ]
+    }
+
+To dynamically modify an RPC method's authorization, use:
+
+.. code-block:: console
+
+    vctl auth rpc add <agent_id.method> <authorized capability 1> <authorized capability 2> ...
```
craig8, line 509: Please provide a real example here...just the template is not enough for this complex of a thing

craig8 (changes requested):

docs/source/platform-features/control/authentication-commands.rst
sgilbride: Removed 'arbitrary'

docs/source/platform-features/control/authentication-commands.rst
sgilbride: Added example for adding and removing an authorization, and the consequences of each.

PR #2700, opened 2021-06-14
Web test update

sgilbride:
Description
-------------
Moved web tests into the web directory.
Added integration tests to test_web_authentication.py for testing RPC calls to AuthService for management of pending ZMQ credentials.
Added check to test_web_authentication.py to verify if the user has a web-enabled VOLTTRON installation before running tests.

Type of change
-----------------
Updated Tests

How Has This Been Tested?
------------------------------
Ran pytest on the web/tests directory using PyCharm IDE. Some of the new integration tests in test_web_authentication.py needed to be re-ran separately before they passed. These were re-ran using the re-run failed tests option from within PyCharm, or by running the tests individually. All tests are passing.


craig8:
These should go into volttrontesting/platform/web

NOT in the main treee

PR #2726, opened 2021-07-20
Web Admin Page Timeout Fix

sgilbride:
# Description

When the access token expires and the admin web page is refreshed, an exception is caused. With this fix, when the access token expires and the page is refreshed, a warning will be stored in the log and the user will be redirected to the login page.

Fixes # (issue)

## Type of change

Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Manual test using a very reduced access token expiration time (30 seconds).
Verified that on a refresh the warning is logged, and the website is reset to the login page.
Ran web pytests and verified all pass.

### DIFF
--- a/volttron/platform/web/admin_endpoints.py
+++ b/volttron/platform/web/admin_endpoints.py
@@ -41,8 +41,6 @@ import os
 import re
 from urllib.parse import parse_qs
 
-import jwt
-
 from volttron.platform.agent.known_identities import PLATFORM_WEB, AUTH
 from volttron.platform.jsonrpc import RemoteError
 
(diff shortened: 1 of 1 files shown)


volttron/platform/web/admin_endpoints.py
```diff
@@ -41,7 +41,10 @@
 import re
 from urllib.parse import parse_qs
 
+import jwt
```
craig8, line 44: why are we adding jwt to this module?

craig8 (changes requested):

volttron/platform/web/admin_endpoints.py
sgilbride: No need. Leftover from a call to jwt.ExpiredSignatureError, which is now handled in the RemoteError.
Removed.

shwethanidd (approved):

PR #2685, opened 2021-05-25
WIP: Add BACnet integ tests

bonicim:
# Description

* Add integration tests (set and get) for BACnet Driver
* Add Dockerfile.test.bacnet, bacnet_device_fixture.py to serve as the "mock" BACnet device for integration test purposes

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

**Test Configuration**:
VM: Ubuntu 18.04
Docker Version: 20.10.1
ZMQ and RMQ installed
Volttron 8.0

Log output from running test: 

```shell
$  pytest services/core/PlatformDriverAgent/tests/test_bacnet_driver.py

===================================================== test session starts =====================================================
platform linux -- Python 3.6.9, pytest-6.2.4, py-1.10.0, pluggy-0.13.1 -- /home/mark2/Workplace/volttron/env/bin/python3
cachedir: .pytest_cache
rootdir: /home/mark2/Workplace/volttron, configfile: pytest.ini
plugins: timeout-1.4.2
timeout: 240.0s
timeout method: signal
timeout func_only: False
collected 2 items                                                                                                             

test_bacnet_driver.py::test_set_and_get[volttron_instance0] PASSED                                                                                                       [ 50%]
test_bacnet_driver.py::test_set_and_get[volttron_instance1] PASSED                                                                                             [100%]


============================================================= 2 passed, 2 warnings in 232.08s (0:03:52) ==============================================================
```

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #2728, opened 2021-07-21
Added handler for devices that do not support use_read_multiple by de…

acedrew:
…fault

# Description

Automatically resolves situation where a device is configured to use_read_multiple but rejects read_multiple requests with the 'unrecognized-service' error.

Fixes # (issue)

## Type of change

- [ x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Had device configured to use_read_multiple that was responding with  rejection errors and deployed code, scrapes happened successfully.

# Checklist:

- [x ] My code follows the style guidelines of this project
- [ x] I have performed a self-review of my own code

### DIFF
--- a/services/core/PlatformDriverAgent/platform_driver/interfaces/bacnet.py
+++ b/services/core/PlatformDriverAgent/platform_driver/interfaces/bacnet.py
@@ -171,6 +171,10 @@ class Interface(BaseInterface):
                     self.max_per_request = max(int(self.register_count/self.register_count_divisor), 1)
                     _log.info("Device requires a lower max_per_request setting. Trying: "+str(self.max_per_request))
                     continue
+                elif e.message.endswith("rejected the request: 9") and self.use_read_multiple:
+                    _log.info("Device rejected request with 'unrecognized-service' error, attempting to access with use_read_multiple false")
+                    self.use_read_multiple = False
+                    continue
                 else:
                     raise
             except errors.Unreachable:


bonicim:
Hi @acedrew, 

What was the device configuration (i.e. bacnet config file) that triggered this RemoteError? I want to see if I can create an integ test for this case as part of my PR #2685  to add Bacnet integ tests, which we currently don't have.

craig8:
@acedrew ping...

acedrew:
It all depends on the end device, I believe it just gave a generic reject with reason '9' which I looked up in the BACnet standard as unrecognized service. The only config requirement is leaving 'use_read_multiple' as the default, 'true' and trying to communicate with a device that does not support read multiple and fails with that specific error code.

bonicim:
> It all depends on the end device, I believe it just gave a generic reject with reason '9' which I looked up in the BACnet standard as unrecognized service. The only config requirement is leaving 'use_read_multiple' as the default, 'true' and trying to communicate with a device that does not support read multiple and fails with that specific error code.

Thanks. I've tested this branch on the BACnet device that we have at the PNNL campus. The Platform Driver and BACnet driver behaved as expected. While this end device was not setup to simulate the conditions required to test this case (i.e. a device that does not support read multiple that fails with error code '9'), I was able to verify with confidence that this additional elseif clause did not negatively affect the behavior of the BACnet driver.

PR #2729, opened 2021-07-31
Fix install_platform_driver_configs script

bonicim:
# Description

install_platform_driver_configs.py does not currently work because it needs the edit_config_store capability. This PR adds that capability, allowing the script to install driver configs. I have also added some checks to ensure that the agent created in the script was successfully created. 

## Type of change

# How Has This Been Tested?

I used the Bacnet scripts to generate a directory containing a driver config file and two directories named 'registry' and 'devices'. I used that directory as the input directory for install_platform_driver_configs.py and tested the script. Steps and output is below:

```
$ cd ~volttron/scripts/
$ python bacnet_scan.py --address <some ip address associated with a Bacnet device> --csv-out test.csv
$ python grab_multiple_configs.py test.csv --ini BACpypes.ini
```

```
$ python install_platform_driver_configs.py /home/mark2/Workplace/volttron/scripts/bacnet/bacconfigs

# Output
Deleting old Platform Driver store
Storing main configuration
Storing configuration: registry_configs/506892.csv
Storing configuration: devices/506892
```
# Checklist:


scripts/install_platform_driver_configs.py
```diff
@@ -76,17 +77,30 @@
...
-
-    keystore = KeyStore()
-    agent = Agent(address=get_address(), identity="platform_driver_update_agent",
-                  publickey=keystore.public, secretkey=keystore.secret,
-                  enable_store=False)
+    ks = KeyStore()
+    publickey = ks.public
+    secretkey = ks.secret
+    identity = "platform_driver_update_agent"
+    capabilities = dict(edit_config_store=dict(identity=PLATFORM_DRIVER))
```
craig8, line 87: I think this would be better to have you use the execute_commands function.  This Modifies the auth file directly and we are trying not to have that be done by any other location than the Auth agent itself.  The creation of the agent can be done using build_agent rathern than the Agent interface.

craig8 (changes requested):

scripts/install_platform_driver_configs.py
bonicim: I used `build_platform_wrapper` and `execute_command` to make this script use those tools to make changes to the Volttron platform instead of changing the internals directly.

scripts/install_platform_driver_configs.py
```diff
@@ -36,15 +36,15 @@
...
-from volttron.platform.auth import AuthFile, AuthEntry
-from volttron.platform.vip.agent import Agent
+import os
+import glob
+
 from volttron.platform.agent.known_identities import CONFIGURATION_STORE, PLATFORM_DRIVER
 from volttron.platform import get_address
 from volttron.platform.keystore import KeyStore
+from volttron.platform.agent.utils import execute_command
+from volttrontesting.fixtures.volttron_platform_fixtures import build_wrapper
```
craig8, line 46: It does not make sense to use platform wrapper in this

craig8 (changes requested):

craig8:
Use build_agent from vip.utils module if you require it.

scripts/install_platform_driver_configs.py
bonicim: Yes, removing build_wrapper and using the build_agent in utils from vip.utils instead.

scripts/install_platform_driver_configs.py
```diff
@@ -39,11 +39,9 @@
...
 import glob
 
-from volttron.platform.agent.known_identities import CONFIGURATION_STORE, PLATFORM_DRIVER
-from volttron.platform import get_address
+from volttron.platform.agent.known_identities import CONFIGURATION_STORE, PLATFORM_DRIVER, PLATFORM
 from volttron.platform.keystore import KeyStore
-from volttron.platform.agent.utils import execute_command
-from volttrontesting.fixtures.volttron_platform_fixtures import build_wrapper
+from volttron.platform.vip.agent.utils import build_agent
 from argparse import ArgumentParser, RawTextHelpFormatter
```
craig8, line 45: Please check import order as argparse should be before any of the volttron items.

craig8 (commented):
```(volttron) (base) gridappsd@gridappsd_dev:/repos/volttron$ python scripts/install_platform_driver_configs.py foo
Traceback (most recent call last):
  File "scripts/install_platform_driver_configs.py", line 133, in <module>
    install_configs(args.input_directory, args.keep_old)
  File "scripts/install_platform_driver_configs.py", line 80, in install_configs
    os.chdir(input_directory)
FileNotFoundError: [Errno 2] No such file or directory: 'foo'
```
should not have the traceback information just the missing directory.

scripts/install_platform_driver_configs.py
bonicim: Will do.

bonicim:
> ```
> Traceback (most recent call last):
>   File "scripts/install_platform_driver_configs.py", line 133, in <module>
>     install_configs(args.input_directory, args.keep_old)
>   File "scripts/install_platform_driver_configs.py", line 80, in install_configs
>     os.chdir(input_directory)
> FileNotFoundError: [Errno 2] No such file or directory: 'foo'
> ```
> 
> should not have the traceback information just the missing directory.

Roger. Will change.

PR #2736, opened 2021-08-07
Revert 2735 remove venv assumption

craig8:
Mistakely merged pr


craig8:
Not going to change this back :(

PR #2745, opened 2021-08-16
Fixed CSR Authorizations.

sgilbride:
# Description

Fixed issue with CSR approval requiring the approval command to be used twice.
This issue was caused by a csr missing some of the required subjects. This led to an unhandled exception causing the csr approval to break after a cert was created. Since the cert was created before the unhandled exception, when an approval request was made again, the cert was detected, and the breaking logic was skipped over, allowing the csr approval to complete without issue on the second attempt.

Also updated some VC tests to match current jsonrpc functionality.

## Type of change

# How Has This Been Tested?

Tested manually using dynamic agents.
Tested with pytests, especially web, auth, and control tests.


craig8 (commented):
Please update documentation for create_csr as there is no name parameter.  Add documentation for the build_subject including return types.

sgilbride:
@craig8 Updated docstrings for create_csr and build_subject

volttron/platform/certs.py
```diff
@@ -503,8 +503,8 @@ def create_csr(self, fully_qualified_identity, remote_instance_name):
         The key that is used to sign the csr is <instance_name>.name.
 
         :param fully_qualified_identity:
-        :param target_volttron:
-        :return:
+        :param remote_instance_name:
+        :return csr.public_bytes:
```
craig8, line 507: So what exactly is csr.publich_bytes...https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization/?highlight=public_bytes#cryptography.hazmat.primitives.serialization.PublicFormat

craig8 (changes requested):

volttron/platform/certs.py
sgilbride: It looks to be an encoded form of the certificate, which can be stored or transmitted. Updated docstring to better reflect this.

craig8 (approved):

shwethanidd (approved):

PR #2748, opened 2021-08-17
Historian  test case fixes 

schandrika:
Historian test case fixes


volttrontesting/services/historian/test_multiplatform.py
```diff
@@ -61,91 +61,6 @@
 from volttron.platform.agent.utils import execute_command
 
 
-@pytest.fixture(scope="module")
```
craig8, line 64: Is this fixture somewhere else now?

PR #2751, opened 2021-08-18
vctl auth serverkey requires volttron to be running.

sgilbride:
# Description

vctl auth parser has some commands that do not require the platform to be running to execute. vctl auth serverkey relies on the platform to be running, but does not check if a connection to the platform is available. This PR is to address that.

Fixes #

## Type of change

# How Has This Been Tested?

Tested using vctl with and without platform running.

### DIFF
--- a/volttron/platform/control.py
+++ b/volttron/platform/control.py
@@ -1735,7 +1735,14 @@ def show_serverkey(opts):
 
     return 0 if success, 1 if false
     """
-    q = Query(opts.connection.server.core)
+    conn = opts.connection
+    if not conn:
+        _stderr.write(
+            "VOLTTRON is not running. This command "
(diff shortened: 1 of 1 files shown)


volttron/platform/control.py
```diff
@@ -1735,7 +1735,14 @@ def show_serverkey(opts):
...
     return 0 if success, 1 if false
     """
-    q = Query(opts.connection.server.core)
+    conn = opts.connection
+    if not conn:
+        _stderr.write(
+            "VOLTTRON is not running. This command "
+            "requires VOLTTRON platform to be running\n"
+        )
+        return
```
craig8, line 1744: as an error this should return 1 or not None or 0

sgilbride: Fixed.

PR #2756, opened 2021-08-20
Add workflows for dbutils tests; update db images for testing

bonicim:
# Description

Creates separate Workflows for each of the test for dbutils except for mongoutils because MongoDBHistorian is unsupported. 

Reduces number of database images to be tested to 2-3 images in accordance with what documentation states on what databases have been tested on. 

## Type of change

# How Has This Been Tested?

Ran Workflows on my personal Github account. See https://github.com/bonicim/volttron/actions

# Checklist:

### DIFF
--- a/.github/workflows/pytest-dbutils-backup_db.yml
+++ b/.github/workflows/pytest-dbutils-backup_db.yml
@@ -1,0 +1,68 @@
+---
+# This workflow is meant as a foundational workflow for running integration/unit tests on multiple targeted
+# ubuntu versions with multiple python versions.
+#
+# This workflow utilizes the build-dependency-cache workflow which sets up the environment dependencies using
+# bootstrap.py --all
+#
+
+# Documentation for the syntax of this file is located
+# https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions
+
+# The workflow name will show up in the action tab on github during execution
+# https://github.com/VOLTTRON/volttron/actions (or if you are pushing to your own fork change the user)
+name: Testing BackupDatabase
+
+# Determine what events are going to trigger a running of the workflow
+on: [pull_request]
+
+jobs:
+  build:
+    env:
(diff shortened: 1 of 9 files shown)


.github/workflows/pytest-dbutils-backup_db.yml
```diff
@@ -0,0 +1,65 @@
...
+# https://github.com/VOLTTRON/volttron/actions (or if you are pushing to your own fork change the user)
+name: Testing BackupDatabase
+
+# Determine what events are going to trigger a running of the workflow
+on: [pull_request]
+
+jobs:
+  build:
+    env:
+        TEST_FILE: volttrontesting/platform/dbutils/test_backup_database.py
```
craig8, line 22: I think this should not include the path...just test_backup_database.py....if you look down at the path output you would have a really long and probably broken file for each of these tests.

.github/workflows/pytest-dbutils-backup_db.yml
```diff
@@ -0,0 +1,65 @@
...
+          python-version: ${{ matrix.python-version }}
+
+      # Run the specified tests and save the results to a unique file that can be archived for later analysis.
+      - name: Run pytest on ${{ matrix.python-version }}, ${{ matrix.os }}
+        uses: volttron/volttron-build-action@v3
+        timeout-minutes: 600
+        with:
+            python_version: ${{ matrix.python-version }}
+            os: ${{ matrix.os }}
+            test_path: ${{ env.TEST_FILE }}
```
craig8, line 55: This would be path to the test and then the TEST_FILE, unless you wanted to add TEST_DIR as another env...

I almost thing the env thing is overkill for these.

craig8 (changes requested):
Same comments throughout.

.github/workflows/pytest-dbutils-backup_db.yml
bonicim: Good catch. I forgot about how that env var was used in that last step. Will change the path to just use the name of the test itself.

PR #2760, opened 2021-08-24
Fixed Agent Install Config in VCFG

sgilbride:
# Description

Vcfg used **vctl install** internally to install agents to the volttron instance with the "--config" option to set the agent configuration.  The "--config" option is used to select the configuration of the volttron instance that vctl is using, not the agent config. This has been corrected to use the correct option for setting the agent's configuration file "--agent-config". This bug was discovered when attempting to install the SQL Platform Historian with vcfg. The agent installed correctly, but was unable to start due to a broken configuration.

## Type of change

# How Has This Been Tested?

Manually tested using vcfg to install the SQL Platform Historian, and verified it was able to start correctly.


shwethanidd:
@sgilbride: can you make another PR to develop?

PR #2785, opened 2021-09-10
Added volttron-update-auth

sgilbride:
# Description

Added volttron-update-auth to upgrade older auth files. This will include the identity field in each auth entry so long as the agent is installed on the platform. The command uses the publickey of the agent to map the identity from the installed agent's directory to the auth entry. This command must be run locally. Once this is done, the agent can use dynamic rpc authorizations.

# Fixes

Allows backwards compatibility for dynamic rpc authorization. 
Prevents the "Auth entry not found for <agent_id>: rpc_method_authorizations not updated" from occurring unexpectedly.

## Type of change

# How Has This Been Tested?

Verified that by running the command, empty/missing identity fields for existing installed agents were added.

### DIFF
--- a/volttron/platform/update_auth_file.py
+++ b/volttron/platform/update_auth_file.py
@@ -35,20 +35,77 @@
 # BATTELLE for the UNITED STATES DEPARTMENT OF ENERGY
 # under Contract DE-AC05-76RL01830
 # }}}
+import os
 import sys
+from shutil import move
+from gevent import monkey as curious_george
+curious_george.patch_all(thread=False, select=False)
 
 from volttron.platform import get_home
 from volttron.platform.aip import AIPplatform
 from volttron.platform.auth import AuthFile
 from volttron.platform.instance_setup import fail_if_instance_running
+from volttron.platform.keystore import KeyStore
 
 
-def get_identity_credentials():
-    """Returns a dictionary containing a mapping from publickey to identity"""
+def get_aip():
+    """Get AIPplatform to interface with agent directories in vhome"""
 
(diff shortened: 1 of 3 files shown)


sgilbride:
Need to add pytests.

craig8:
Lots of monkey patch warnings when running this script.

```
(volttron) d3m614@8:~/volttron-8.1.1$ volttron-update-auth                                    │
/home/d3m614/volttron-8.1.1/env/lib/python3.6/site-packages/grequests.py:22: MonkeyPatchWarnin│
g: Monkey-patching ssl after ssl has already been imported may lead to errors, including Recur│
sionError on Python 3.6. It may also silently lead to incorrect behaviour on Python 3.7. Pleas│
e monkey-patch earlier. See https://github.com/gevent/gevent/issues/1016. Modules that had dir│
ect imports (NOT patched): ['urllib3.contrib.pyopenssl (/home/d3m614/volttron-8.1.1/env/lib/py│
thon3.6/site-packages/urllib3/contrib/pyopenssl.py)'].                                        │
  curious_george.patch_all(thread=False, select=False)                                        │
Auth File Update Complete!     
```

craig8:
Where are we going to add documentation for this?  Add an Upgrading section to the README or in RTD? @sgilbride @shwethanidd

shwethanidd:
I would say both.

schandrika:
> I would say both.

@sgilbride Where is the documentation regarding this auth update? I would like to get started with testing.

craig8:
Fixes #2790 documentation is in #2787

PR #2787, opened 2021-09-15
Added VOLTTRON 8.x Update Message to README

sgilbride:
# Description

Added note about required steps to upgrade to version 8.x, by running volttron-update-auth.
This was added to the README and the upgrading-version.rst doc file.
Updated test_auth_control.py to be more robust.

## Type of change

- [ ] Updated documentation.
- [ ] Updated auth test.

# How Has This Been Tested?

Ran test_auth_control.py locally twice to verify passing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,6 +296,13 @@ There are several walkthroughs to explore additional aspects of the platform:
 -   [RabbitMQ setup with Federation and Shovel plugins](https://volttron.readthedocs.io/en/latest/deploying-volttron/multi-platform/multi-platform-rabbitmq-deployment.html)
 -   [Backward compatibility with the RabbitMQ message bus](https://volttron.readthedocs.io/en/latest/deploying-volttron/multi-platform/multi-platform-multi-bus.html)
 
+## Upgrade to VOLTTRON 8.x
+
+VOLTTRON 8 and above introduces dynamic RPC authorization, which requires a modification to the auth file.
+If you have a pre-existing instance of VOLTTRON running on an older version, the auth file will need to be updated. 
+This can be done by running the ```volttron-update-auth``` in an activated VOLTTRON environment. The platform can not 
+be running when this update script is ran. 
+
 ## Acquiring Third Party Agent Code
 
 Third party agents are available under the volttron-applications repository. In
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -296,6 +296,13 @@ There are several walkthroughs to explore additional aspects of the platform:
 -   [RabbitMQ setup with Federation and Shovel plugins](https://volttron.readthedocs.io/en/latest/deploying-volttron/multi-platform/multi-platform-rabbitmq-deployment.html)
 -   [Backward compatibility with the RabbitMQ message bus](https://volttron.readthedocs.io/en/latest/deploying-volttron/multi-platform/multi-platform-multi-bus.html)
 
+## Upgrade to VOLTTRON 8.x
```
craig8, line 299: @shwethanidd and @schandrika 
Lets put this at the top of the readme hoping people will see it.  Call it Upgrading rather than upgrade.

volttrontesting/platform/auth_tests/test_auth_control.py
```diff
@@ -274,16 +274,16 @@ def test_auth_remove(auth_instance):
...
+        gevent.sleep(i)
         entries = auth_list_json(platform)
         i += 1
     auth_add(platform, _auth_entry6)
     entries = auth_list_json(platform)
     assert len(entries) > 0
     i = 0
-    while len(entries) < (len_entries + 1) and i < auth_retry:
-        gevent.sleep(1)
+    while len(entries) <= (len_entries + 1) and i < auth_retry:
```
craig8, line 285: why is this one len_entries + 1 while the others are len_entries?
why are we changing these?

craig8 (changes requested):
The user will need to re-run bootstrap.py in order to do the upgrade.  Please add how to upgrade to the documentation.  I have also seen that the volttron-upgrade-auth sometimes doesn't register if you have an existing install, so the user would have to remove the environment and the volttron.egg-info directories before bootstrapping again.

volttrontesting/platform/auth_tests/test_auth_control.py
schandrika: because there are 2 auth_add in the test case at this point and len_entries is length before any add by test case

README.md
sgilbride: Updated!

sgilbride: Included steps needed for bootstrap.py.

PR #2795, opened 2021-09-20
Removing duplicate package entries in requirements.py and bootstrap.py

schandrika:
fix for jira-533 - Removing duplicate package entries in requirements.py. 
moved wheel to list of requirements in requirements.py instead of explicit install from bootstrap.py. 
Pinned version of pyzmq to 22.2.1 as current version(22.3.1) does not install correctly

Tested by bootstrapping for zmq and rmq and did a sanity test that includes
1. start and stop platform
2. vctl install listener
3. vcfg rabittmq single
4. vctl install sqlite historian


craig8:
fix will be routed from #2797

PR #2825, opened 2021-11-17
Bugfix/control connection check

sgilbride:
# Description

Provides a check if the platform is running for control methods.
Adds checks to appropriate methods so vctl will behave correctly in cases where the platform is or is not running.

## Type of change

# How Has This Been Tested?

Tested with pytest locally running control_tests test directory.

### DIFF
--- a/volttron/platform/control.py
+++ b/volttron/platform/control.py
@@ -794,11 +794,16 @@ Agent = collections.namedtuple("Agent",
                                "name tag uuid vip_identity agent_user")
 
 
-def needs_connection(func):
+def needs_connection(func: callable) -> callable:
     """
     Decorator that checks if a connection exists.
     If opts.connection exists, then the platform is running.
     Used for commands that need to connect to the platform.
+    If opts.connection does not exist, then an error message will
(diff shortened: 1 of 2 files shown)


volttron/platform/control.py
```diff
@@ -1728,20 +1742,20 @@ def do_stats(opts):
...
     return 0 if success, 1 if false
     """
     conn = opts.connection
-    if not conn:
-        _stderr.write(
-            "VOLTTRON is not running. This command "
-            "requires VOLTTRON platform to be running\n"
-        )
-        return 1
+    # if not conn:
```
craig8, line 1753: Remove commented out code..

craig8 (changes requested):
I like this decorator.

The documentation for needs_connection should specify that opts must be the first argument to func which is passed in and that the connection must exist...

So document the arguments to this as well as the func should be :callable I believe.

volttron/platform/control.py
sgilbride: Fixed!

craig8 (approved):

PR #2831, opened 2021-11-29
return subprocess returncodes

### LINKED ISSUE #2829: install_agent.py does not send a failure code from vctl subcommand
NOTE issues are for bugs/feature requests not support.  Please ask at https://stackoverflow.com/questions/tagged/volttron for help

### Description of Issue
If there is an error (and error return code in the shell), the install_agent.py script does not provide that code back to its own return, and instead returns 0 (success) (https://github.com/VOLTTRON/volttron/blob/releases/8.x/scripts/install-agent.py#L39)
This line needs to be `sys.exit(output)` to return the same code (or should process the code and return a non-zero code).

laroque:
# Description

Update to install-agents script to ensure return codes from subprocesses make it back to the script return. Should resolve #2829

### DIFF
--- a/scripts/install-agent.py
+++ b/scripts/install-agent.py
@@ -36,7 +36,7 @@ if not ignore_env_check and not inenv and not corrected:
     cmds.extend(sys.argv[1:])
     try:
         output = subprocess.call(cmds, env=os.environ)
-        sys.exit(0)
+        sys.exit(output)
     except RuntimeError:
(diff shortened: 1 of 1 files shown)


laroque:
I tested this on my local py with and without these changes and confirmed that `install-agents.py -s <some-bogus-path>` was returning 0 with a failure message printed before, and returns an error code now.

Issue #2821, opened 2021-10-21
caching metadata on SQLhistorian start fails when metadata contains NULL


schandrika:
caching metadata on SQLhistorian start fails when metadata contains NULL
Affected versions- volttron 8.x and develop

bbartling:
We are running into this issue when we updated our Central instance to version 8, same thing? Here's a snip from the log file:

```
2022-05-20 23:51:07,648 (sqlhistorianagent-4.0.0 1993611) volttron.platform.agent.base_historian INFO: Trying to setup historian
2022-05-20 23:51:07,648 (sqlhistorianagent-4.0.0 1993611) __main__ INFO: historian_setup on Thread: Thread-1
2022-05-20 23:51:07,701 (sqlhistorianagent-4.0.0 1993611) volttron.platform.agent.base_historian ERROR: Failed to setup historian!
Traceback (most recent call last):
  File "/var/lib/volttron/volttron/platform/agent/base_historian.py", line 1292, in _historian_setup
    self.historian_setup()
  File "/var/lib/volttron/volttron/utils/docs.py", line 47, in f
    return self.mthd(obj, *args, **kwargs)
  File "/home/volttron/.volttron/agents/06de9128-a5a8-41eb-ac7c-1e7821653ea8/sqlhistorianagent-4.0.0/sqlhistorian/historian.py", line 352, in historian_setup
    topic_meta_map = self.bg_thread_dbutils.get_topic_meta_map()
  File "/var/lib/volttron/volttron/platform/dbutils/postgresqlfuncts.py", line 352, in get_topic_meta_map
    rows = self.select(query)
  File "/var/lib/volttron/volttron/platform/dbutils/basedb.py", line 436, in select
    cursor.execute(query, args)
psycopg2.errors.UndefinedColumn: column "metadata" does not exist
LINE 1: SELECT topic_id, metadata FROM "topics"
```

PR #2832, opened 2021-12-02
Hotfix - sqlhistorian

bonicim:
* Hotfix to address #2821, which was resolved in PR #2822 
* This PR is a cherry pick of #2822 onto the 'main' branch


craig8:
This should not be merged into that branch .... this should be merged into the 8.x branch and hot fix from there. (note this is different than "normal" the 8.1.1 is fixed point in time.

PR #2839, opened 2021-12-15
Upgrading to new rabbitmq server and pika

### LINKED ISSUE #2703: RabbitMQ VOLTTRON does not work in Ubuntu 20.04 LTS
RabbitMQ works with following configurations:

erlang: 1.23.1
rabbitmq-server: 3.8.9
pika: 1.1.0

There has been significant changes in RabbitMQ and Pika in the past 2 years. It is not easy to create gevent adapter with latest version of pika. Instead, we need to update VOLTTRON  such that gevents and async work co-cooperatively. After that we can investigate replacing pika's gevent adapter with asyncio adapter.
 
### Affected Version
8.x

shwethanidd:
# Description

Pika 1.2.0 now supports GeventAdapter natively. With this we can use latest version of rabbitmq (3.9.7) and update supporting packages such as cryptography. This PR covers pika, rabbitmq and crypto updates.

Fixes #2703


shwethanidd:
@schandrika Can you test this?

volttron/platform/vip/agent/core.py
```diff
@@ -1008,6 +1007,7 @@ def connect_callback():
                             binding['routing_key'] == router_key:
                         router_connected = True
                         break
+            router_connected = True
```
schandrika, line 1010: Earlier we set router_connected=True only if there are bindings.  but it is outside of "if bindings:" so wont the flag get set even if there was attribute error earlier?

volttron/platform/vip/proxy_zmq_router.py
```diff
@@ -97,57 +97,57 @@ def startup(self, sender, **kwargs):
...
+        channel.queue_bind(self._rpc_handler_queue,
+                           connection.exchange,
                            routing_key=self.core.instance_name + '.proxy.router.zmq.outbound.subsystem',
                            callback=None)
-        channel.basic_consume(self.rpc_message_handler,
-                              queue=self._rpc_handler_queue,
-                              no_ack=True)
+        channel.basic_consume(self._rpc_handler_queue,
+                              self.rpc_message_handler
+                              )
```
schandrika, line 111: if you remove no_ack=True, do you require a channel.basic_ack(method.delivery_tag) like the one you added in rmq_pubsub.py ?

schandrika:
test_federation.py fails with gevent error

2021-12-20 16:39:05,653 pika.callback ERROR: Calling <bound method RMQConnection.on_open_error of <volttron.platform.vip.rmq_connection.RMQConnection object at 0x7f394f9b5b80>> for "0:_on_connection_error" failed
gevent.exceptions.BlockingSwitchOutError: Impossible to call blocking function in the event loop callback

schandrika:
All basic operations from command line work for rmq single instance. Test cases and multiplatform use cases still need fixing but merging this into develop so that we have a newer version of rabbitmq, pika, and cryptography version in develop branch for other agent's use.

schandrika:
Created Issue #2864

PR #2850, opened 2021-12-30
Set pytestmark for Ambient tests

bonicim:
# Description

Fixes #2849 

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [x] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #2857, opened 2022-01-21
Create workflow to test miscellaneous platform tests

gwenkidd:
# Description

Created a new workflow to be added into our CI pipeline that will test miscellaneous platform tests. 

Fixes #VOLTDEV-572

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

This workflow runs platform tests.

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain:
* SDK:

# Checklist:


.github/workflows/pytest-miscellaneous-tests.yml
```diff
@@ -0,0 +1,102 @@
...
+      - uses: actions/checkout@v2
+
+      # Setup the python environment for the operating system
+      - name: Set up Python ${{matrix.os}} ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      # Run the specified tests and save the results to a unique file that can be archived for later analysis
+      - name: Run certs test on ${{ matrix.python-version }}, ${{ matrix.os }}
```
craig8, line 56: Is there a reason we are specifying exactly the tests to run rather than running the directory of tests?

if test_path was volttrontesting/platform that would get all of the tests here and if there are any additional the script would automatically pick them up and test them.

gwenkidd: @bonicim had told me to only include those tests in /platform that are passing. There were several that were resulting in errors or failing that he said not to include, and we would work on getting those fixed in a different task. So rather than testing the whole directory, I just listed the ones that were passing.

bonicim: @gwenkidd Craig made a good point about what to do with the failing tests. Consequently, let's not only document the failing tests as Jira tickets but also update those failing tests using pytests markers: see the updated Jira ticket for details. More importantly, this will increase the scope of the Jira ticket and thus this PR. Please put this PR in draft mode see https://github.blog/changelog/2020-04-08-convert-pull-request-to-draft/; then mark it as ready for review when it's ready.
@gwenkidd Craig made a good point about what to do with the failing tests. Consequently, let's not only document the failing tests as Jira tickets but also update those failing tests using pytests markers: see the updated Jira ticket for details. More importantly, this will increase the scope of the Jira ticket and thus this PR. Please put this PR in draft mode see https://github.blog/changelog/2020-04-08-convert-pull-request-to-draft/; then mark it as ready for review when it's ready.

gwenkidd: @bonicim Will do!

gwenkidd: @bonicim All the failing/erroring tests have been marked with the pytest fail marker.

volttrontesting/platform/test_connection.py
```diff
@@ -40,13 +40,15 @@ def setup_control_connection(request, get_volttron_instances):
 
 
 @pytest.mark.control
+@pytest.mark.xfail
```
bonicim, line 43: If you want to mark that **all** the tests as expected to fail, you can mark the test file itself as expected to fail instead of having to mark every individual test. In other words, you can declare a pytestmark global variable and then define it based on whatever pytest.marks you want and they will apply to all the tests. 

```python

import pytest

pytestmark = [pytest.mark.control, pytest.mark.xfail]
``` 

And see this PR as an example in Volttron: #2850 

Ref: https://docs.pytest.org/en/latest/reference/reference.html#globalvar-pytestmark

bonicim (changes requested):

volttrontesting/platform/test_connection.py
craig8: I think we should not use global marks for this....because when we fix one of them we have to put the marks back on the other functions...

gwenkidd:
> Why are we doing this globally?

gwenkidd:
> Why are we doing this globally?

Mark had mentioned to. I can change it back.

craig8:
@gwenkidd I didn't see the conversation just the diff when I made that remark.

gwenkidd:
> @gwenkidd I didn't see the conversation just the diff when I made that remark.

Ah okay. I would agree that having the markers on each test would make more sense for when we go in to fix them.

gwenkidd:
> > > @gwenkidd I didn't see the conversation just the diff when I made that remark.
> > 
> > 
> > Ah okay. I would agree that having the markers on each test would make more sense for when we go in to fix them.
> 
> @gwenkidd I made the comment about pytest global markers to start a thought process on how to handle failing tests at different levels (i.e. module or function) while trying to keep the tests at a a reasonable balance of [WET and DRY code](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). Craig's comment on having markers at every test makes sense if we're fixing tests one or a subset of the tests at a time, though my preference is that if we are fixing tests, just fix them all at once.

@craig8 so what are we thinking is best then?

craig8:
I think the best is to allocate some resources to fix these tests.  There is no reason the connection tests especially for zmq should be failing.   

For now mark them as xfail globally like you have them and then lets see if we can get a process to actually pass in our environments as well as on gh.

bonicim:
Lgtm. Approved.

PR #2871, opened 2022-02-04
Cert creation unit tests

### LINKED ISSUE #24: Multiple logging handlers may be instatiated with platform.agent.utils.setup_logging()

gwenkidd:
# Description

DRAFT PR: Added unit tests for cert functions. Still need to work on some issues within two tests. 

Fixes #24

## Type of change

# How Has This Been Tested?

These are tests.

# Checklist:

### DIFF
--- a/volttrontesting/platform/test_certs.py
+++ b/volttrontesting/platform/test_certs.py
@@ -1,10 +1,12 @@
 import os
 import tempfile
 import shutil
-
 import pytest
+import json
+from pathlib import Path
 from volttron.platform.certs import Certs, Subject, CertError
 from volttron.platform.agent.utils import get_platform_instance_name
+from volttrontesting.platform.test_certs_utils import *
 
 INSTANCE_NAME = "VC"
 PLATFORM_CONFIG = """
@@ -66,6 +68,7 @@ def temp_volttron_home(request):
     """
     dirpath = tempfile.mkdtemp()
     os.environ['VOLTTRON_HOME'] = dirpath
+    debug_flag = os.environ.get('DEBUG', True)
     with open(os.path.join(dirpath, "platform_config.yml"), 'w') as fp:
         fp.write(PLATFORM_CONFIG)
 
@@ -73,8 +76,36 @@ def temp_volttron_home(request):
         fp.write("[volttron]\n")
         fp.write("instance-name = {}\n".format(INSTANCE_NAME))
     yield dirpath
-    shutil.rmtree(dirpath, ignore_errors=True)
-    assert not os.path.exists(dirpath)
+
+    if not debug_flag:
+        shutil.rmtree(dirpath, ignore_errors=True)
+        assert not os.path.exists(dirpath)
+
+    #shutil.rmtree(dirpath, ignore_errors=True)
+    #assert not os.path.exists(dirpath)
+
+
+@pytest.fixture(scope="function")
+def temp_csr(request):
+    """
+    Create a Certificate Signing Request (CSR) using the Certs class.
+    Use this CSR to test approving, denying, and deleting CSRs
+    """
+    # Create CSR with Volttron code
+    certs = Certs()
+    data = {'C': 'US',
+            'ST': 'Washington',
+            'L': 'Richland',
+            'O': 'pnnl',
+            'OU': 'volttron',
+            'CN': INSTANCE_NAME+"_root_ca"}
+    certs.create_root_ca(**data)
+    assert certs.ca_exists()
+
+    certs.create_signed_cert_files(name="FullyQualifiedIdentity", ca_name=certs.root_ca_name)
+
+    csr = certs.create_csr("FullyQualifiedIdentity", "RemoteInstanceName")
+    yield certs, csr
 
 
 def test_certificate_directories(temp_volttron_home):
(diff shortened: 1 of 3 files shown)


volttrontesting/platform/test_certs_utils.py
```diff
@@ -0,0 +1,236 @@
...
+from typing import Union
+
+PathStr = Union[Path, str]
+
+__all__ = ['TLSRepository']
+
+_log = logging.getLogger(__name__)
+
+
+def execute_command(cmds, env=None, cwd=None, logger=None, err_prefix=None) -> str:
```
craig8, line 15: just use this from the volttron directories...rather than copying it here.

this should be in one of the utils directories do a search for def execute_command and it should come up.  Note from the other repo I didn't have access to volttron's namespace.

craig8 (changes requested):

craig8:
@sgilbride @gwenkidd should we add these to the web tests?  That's where we would see this used the most.  What's your thoughts?

gwenkidd:
> @sgilbride @gwenkidd should we add these to the web tests? That's where we would see this used the most. What's your thoughts?

Pushed all the changes you requested. I believe @sgilbride is still going to take a look at the failing tests in this PR.

volttrontesting/platform/test_certs.py
```diff
@@ -1,10 +1,12 @@
 import os
 import tempfile
 import shutil
-
 import pytest
+import json
+from pathlib import Path
 from volttron.platform.certs import Certs, Subject, CertError
 from volttron.platform.agent.utils import get_platform_instance_name
+from volttrontesting.platform.test_certs_utils import *
```
craig8: perfer not to import * anywhere but in __init__.py files.  better to use explicit import

volttrontesting/platform/test_certs.py
```diff
@@ -66,15 +68,43 @@ def temp_volttron_home(request):
...
+            'O': 'pnnl',
+            'OU': 'volttron',
+            'CN': INSTANCE_NAME+"_root_ca"}
+    certs.create_root_ca(**data)
+    assert certs.ca_exists()
+
+    certs.create_signed_cert_files(name="FullyQualifiedIdentity", ca_name=certs.root_ca_name)
+
+    csr = certs.create_csr("FullyQualifiedIdentity", "RemoteInstanceName")
+    yield certs, csr
```
craig8: Should cleanup after the yield?

Teardown of test?

volttrontesting/platform/test_certs.py
```diff
@@ -66,15 +68,43 @@ def temp_volttron_home(request):
     """
     dirpath = tempfile.mkdtemp()
     os.environ['VOLTTRON_HOME'] = dirpath
+    debug_flag = os.environ.get('DEBUG', True)
```
craig8: Probably should change this method to use the https://github.com/VOLTTRON/volttron/blob/be8668c64d16522a8f6da95924e4f3a14fd03ba7/volttrontesting/utils/platformwrapper.py#L181

craig8 (changes requested):

volttrontesting/platform/web/test_certs.py
```diff
@@ -0,0 +1,284 @@
...
+ssl: 'true'
+
+# defaults to ~/rabbitmq_server/rabbbitmq_server-3.9.7
+rmq-home: "~/rabbitmq_server/rabbitmq_server-3.9.7"
+"""
+
+
+@pytest.fixture(scope="function")
+def temp_volttron_home(request):
+    """
```
craig8, line 64: Should utilize this method inside this function.

https://github.com/VOLTTRON/volttron/blob/be8668c64d16522a8f6da95924e4f3a14fd03ba7/volttrontesting/utils/platformwrapper.py#L181

volttrontesting/platform/web/test_certs.py
```diff
@@ -0,0 +1,284 @@
...
+            'OU': 'volttron',
+            'CN': INSTANCE_NAME+"_root_ca"}
+    certs.create_root_ca(**data)
+    assert certs.ca_exists()
+
+    certs.create_signed_cert_files(name="FullyQualifiedIdentity", ca_name=certs.root_ca_name)
+
+    csr = certs.create_csr("FullyQualifiedIdentity", "RemoteInstanceName")
+    yield certs, csr
+
```
craig8, line 108: Clean up delete certs? csr?

craig8 (changes requested):

volttrontesting/platform/web/test_certs.py
```diff
@@ -155,6 +155,8 @@ def test_create_signed_cert_files(temp_volttron_home):
     assert existing_cert[0] == certs.cert("test_cert")
 
 
+@pytest.mark.skipif('OpenSSL' not in os.listdir("/home/volttron/git/myvolttron/env/lib/python3.8/site-packages"),
```
craig8, line 158: hmmm...we can't make this assumption about where openssl livs.

volttrontesting/platform/web/test_certs.py
craig8: I meant the actual files not the variables.

volttrontesting/platform/web/test_certs.py
gwenkidd: yea I was a little unsure about this part, do you have a recommendation on how I could check for openssl better?

craig8: A couple of options to determine if you have a package installed:
- use https://docs.python.org/3.8/library/importlib.html
- attempt to import and set a flag to tell you whether or not it is imported
try:
  import openssl
  HAS_OPENSSL = True
except ImportError:
  HAS_OPENSSL = False

craig8:
@gwenkidd the test_certs_utils doesn't have any tests in it.  Please move this file under volttrontesting/utils and change it's name to be certs_utils or something similar.

craig8:
@gwenkidd  the issue #24 does not seem to apply to this pr.

gwenkidd:
> @gwenkidd the issue #24 does not seem to apply to this pr.

Oops I meant the Jira ticket VOLTDEV-24. Can I link Jira tickets to this PR?

craig8:
@gwenkidd  you can but it doesn't make sense to

PR #2875, opened 2022-02-07
RMQ Installation in VCFG

sgilbride:
# Description

Updated VCFG to install rabbitmq dependencies if required when setting up RMQ VOLTTRON.
Uses scripts/rabbit_dependencies.sh to handle dependency set up.

## Type of change

# How Has This Been Tested?

Tested manually. Ran test_instance_setup.py tests to verify tests still worked correctly.


schandrika:
@sgilbride Regarding the question about is_rabbitmq_available() that method is being used just to import pika without error. So I think that method is sufficient. However, this questions brings up a related issue. If we bootstrap with rmq, manually create a config file with messagebus=rmq and start volttron without doing a "vcfg rabbitmq single" would volttron exit with the right error message? Could you test this and see if this a quick fix. If not I can create a separate issue for this

PR #2907, opened 2022-03-04
Use Python3 super() calls

Siecje:
There are many more. I'm not sure if this is worth reviewing or if a core contributor would like to make these changes or if they should be left for now and use the new syntax when creating new methods.


awoz:
Hi all. I understand that super() default parameters are utilized in this pr. Are there any project guidelines or thoughts about code readability in this case?

craig8:
@awoz I am interested in which you find more readable?

craig8:
@shwethanidd  thoughts?

awoz:
A coding style that clearly shows a parameter list is more human readable, and is better suited for IDEs like visual studio code for indexing and debugging class objects. While most IDEs expand objects while debugging, visual source code reviews benefit the most.

shwethanidd:
I agree with @awoz, that's my preference as well.

craig8:
@Siecje I think we are not going to merge this one in.

Issue #2903, opened 2022-03-02
Add Security Vulnerabilities scan as part of CI pipeline


bonicim:
### Description of Issue
As part of the recommendations on security from the SSC team, add Checkmarx action to current CI pipeline so that we can scan for software vulnerabilities in the Volttron source code. Note that I am waiting on the SSC team to give me a Checkmarx account. 

https://github.com/marketplace/actions/checkmarx-cxflow-action
https://stackoverflow.com/questions/42424678/github-integration-with-checkmarx

### Affected Version

8.2

craig8:
@bonicim can we actually do this from outside the lab?

bonicim:
> @bonicim can we actually do this from outside the lab?

I don't know. The impression that I got from SSC team was that once they get me a Checkmarx account, I should be able to hit the endpoint from Github.

PR #2908, opened 2022-03-04
Added code analysis for codeql to the repository

craig8:


craig8:
#2903

craig8:
@bonicim would you take a look at the codeql output?  Just to see whether this is a good thing or no.

bonicim:
@craig8 I took a look at the output that is shown in the Commits tab of this PR. It's very helpful and the messages clearly state what the problem is (e.g. potentially logging sensitive information).  I think this is a good tool to have.

PR #2912, opened 2022-03-07
Remove import * imports

Siecje:


craig8:
@Siecje this pr made all of our tests fail.  I imagine there are some imports that haven't been propagated into the code.

Siecje:
Yeah, the problem with `import *` is that modules can import through modules that have `import *`.

PR #2917, opened 2022-03-10
Fix exception order

Siecje:


awoz:
Just curious, this fix gives remote errors first priority for proper logging?

Siecje:
@awoz Yes, since `RemoteError` inherits `Exception` the `except Exception` code would handle the exception and the exception handling of `RemoteError` would never happen.

awoz:
> 

Forgive me for all the questions. How does one determine RemoteError inheritance and would it be simpler if the Exception logice checked for RemoteError internally, removing the need for exception ordering. thx

Siecje:
Here is an example that you can try in the REPL.
```Python
>>> try:
...   raise ValueError()
... except Exception:
...   print("here")
... except ValueError:
...   print("not here")
...
here
```

> How does one determine RemoteError inheritance

You can look at the definition of `RemoteError`. You can `Ctrl+left click` or `Cmd+left click` it in your IDE to go to the definition or search all files for "class RemoteError".

The definition is 
`class RemoteError(Exception):` which means that `RemoteError` is a child class of `Exception`.

The Pythonic way of handling different exceptions different is to have separate except clauses. That is what Python programmer expect.

But you could capture `Exception as exc` and then compare with `isintance(exc, ValueError)`.

awoz:
Cool. I'm most curious about coding style for readability/maintainability rather than language semantics. Will checkout isintance() though. Thanks for your advice.

Issue #2945, opened 2022-03-29
basehistorian's configure() set max_time_publishing and history_limit_days to timedelta instead of float


schandrika:
basehistorian's configure() set max_time_publishing and history_limit_days to timedelta instead of float.
init and methods that use these parameters expect it to float and not timedelta

craig8:
Can you provide your thoughts on what action you would like to take?  Either float or timedelta, but consistent. Or a this should do this rather than one it does?

@schandrika

schandrika:
I think it should be float. That's what is in the documentation and that is how init() sets it. Config store, and hence configure() method was added later. So to keep it consistent with older version and docs it should be float

craig8:
I agree make it a float so we should be just casting the timedelta.

PR #2946, opened 2022-03-31
Fix test_base_historian_agent_should_filter_duplicates in testutils 

gwenkidd:
# Description
Updated the expected_to_publish_list to be the correct to_publish_list and updated the comment in the test to accurately describe the BaseHistorian publishing process. 

## Type of change
Fixing a test

# How Has This Been Tested?
Got this test to pass. 

# Checklist:

### DIFF
--- a/volttrontesting/testutils/test_base_historian_unit.py
+++ b/volttrontesting/testutils/test_base_historian_unit.py
@@ -41,12 +41,12 @@ def test_base_historian_agent_should_filter_duplicates(base_historian_agent):
         peer=None,
         sender=None,
         bus=None,
-        topic=f"unique_record_topic_1",
+        topic="unique_record_topic_1",
         headers={
-            "Date": f"2020-11-17 21:21:10.189393+00:00",
-            "TimeStamp": f"2020-11-17 21:21:10.189393+00:00",
+            "Date": "2020-11-17 21:21:10.189393+00:00",
+            "TimeStamp": "2020-11-17 21:21:10.189393+00:00",
         },
-        message=f"unique_record_1",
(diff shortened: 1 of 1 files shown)


volttrontesting/testutils/test_base_historian_unit.py
```diff
@@ -36,19 +36,18 @@ def test_base_historian_agent_should_filter_duplicates(base_historian_agent):
...
-            bus=None,
-            topic=f"unique_record_topic{num}",
-            headers={
-                "Date": f"2020-11-17 21:2{num}:10.189393+00:00",
-                "TimeStamp": f"2020-11-17 21:2{num}:10.189393+00:00",
-            },
-            message=f"unique_record_{num}",
-        )
+    # Add a unique record to queue
+    base_historian_agent._capture_record_data(
```
craig8, line 40: Are you sure we want tor remove the f-string here?

gwenkidd: I updated the test to only add one unique item rather than multiple so the expected_to_publish_list wasn't too long, just so the test was a little easier to follow.

craig8: The thing is we generally see a large number of duplicates when we have them so we should probably test for many of them not just single...the timestamp resolution should still only give us one.

gwenkidd: We are testing for multiple duplicates. The first for loop adds three duplicate items. Then we add one unique item after that.

schandrika:
The changes in this PR are not the right solution for this issue. See issue #2950 and #2945

PR #2952, opened 2022-04-13
AGENT_CONFIG not existing should be an error

### LINKED ISSUE #2847: AGENT_CONFIG not existing should be an error
### Description of Issue

When the variable passed to load_config is None that should be fine, since we don't have to set a path for AGENT_CONFIG.  However when the variable AGENT_CONFIG is set there should be an error thrown if it doesn't exist.

In the following code

```python

# volttron/platform/agent/utils.py

def load_config(config_path):
    """Load a JSON-encoded configuration file."""
    if config_path is None:
        _log.info("AGENT_CONFIG does not exist in environment. load_config returning empty configuration.")
        return {}

    ########################  Change Here ##############################
    if not os.path.exists(config_path):
        _log.info("Config file specified by AGENT_CONFIG does not exist. load_config returning empty configuration.")
        return {}

```

An exception should be raised and handled by the caller agent.

gwenkidd:
# Description
Added a ValueError that is raised in load_config when the config_path does not exist. 

Fixes #2847 

## Type of change

# How Has This Been Tested?
Tried installing some example agents with an agent config path that does not exist and was given an error. The error that is displayed is thrown in the install_agent_directory function which runs before we reach the load_config function so we don't see the load_config error. 

# Checklist:

### DIFF
--- a/volttron/platform/agent/utils.py
+++ b/volttron/platform/agent/utils.py
@@ -149,8 +149,7 @@ def load_config(config_path):
         return {}
 
     if not os.path.exists(config_path):
-        _log.info("Config file specified by AGENT_CONFIG does not exist. load_config returning empty configuration.")
-        return {}
+        raise ValueError(f"Config file specified by AGENT_CONFIG path {config_path} does not exist.")
 
     # First attempt parsing the file with a yaml parser (allows comments natively)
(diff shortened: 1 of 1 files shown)


gwenkidd:
Closes #2847

PR #2967, opened 2022-05-13
get rid of the repetivie driver_type: fakedriver

kefeimo:
# Description

In original file
image

Fixes # (issue)

## Type of change

Hot fix

# How Has This Been Tested?

Not applicable

# Checklist:

Not applicable

### DIFF
--- a/examples/configurations/drivers/fake.config
+++ b/examples/configurations/drivers/fake.config
@@ -11,5 +11,4 @@
     "campus": "campus",
     "building": "building",
     "unit": "fake_device",
-    "driver_type": "fakedriver"
 }
--- a/examples/configurations/drivers/fake.config
+++ b/examples/configurations/drivers/fake.config
@@ -10,5 +10,5 @@
     "publish_breadth_first": false,
     "campus": "campus",
     "building": "building",
-    "unit": "fake_device",
+    "unit": "fake_device"
 }


awoz:
Change looks fine although the preceding line 'comma' should also be removed.

I also wonder if the underlying read method for this file has an inherent bug. If this is the case, then read logic may want to reject duplicate entries and report an error. Secure coding standards typically require that system 'input' parameters be validated to eliminate potential vulnerabilities. In this regard, consider improving the underlying read logic for this config file.

craig8:
@awoz You are right about the secure coding standard.  Right now we kinda flub this, because we have comments in the json file and duplicates get removed before storage into the config store.  

I wouldn't mind seeing these get turned into yaml files so we don't have to preprocess stupid comments...but sigh.

kefeimo:
@awoz Thanks for the command Andrew, I wrongly set *.config file as *.yaml, so Pycharm didn't catch the syntax error. Thanks for the reference. I am working on a clearer and maybe more secure public interface for driver development. And @craig8, I will use yaml to wrap the current json + csv config. 

image

craig8:
@kefeimo Don't change the config to yaml...the configstore doesn't have yaml support right now.  If you would like you can make an issue, but don't develop it for now.

kefeimo:
> @kefeimo Don't change the config to yaml...the configstore doesn't have yaml support right now. If you would like you can make an issue, but don't develop it for now.

I am going to use a facade/adapter for the user to use yaml and generate csv + json. The current code base still interfere with the csv + json.

craig8:
> > @kefeimo Don't change the config to yaml...the configstore doesn't have yaml support right now. If you would like you can make an issue, but don't develop it for now.
> 
> I was planning to use a facade/adapter for the user to use yaml and generate csv + json. The current code base still interact with the csv + json. But will create an issue.

Let's talk on monday

PR #2985, opened 2022-08-15
Added toctree link to web-api documentation.

davidraker:
# Description

Added link in index.rst for web-api documentation.

Fixes # (issue)
#2983

## Type of change

- [ ] documentation

# How Has This Been Tested?

make html, then pull up in browser. Currently, this has the top-level link to the web-api documentation, but does not expand the menu on the left with sub-topics.

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain:
* SDK:

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ with that data.
 
 ## Upgrading to VOLTTRON 8.x
 
-VOLTTRON 8 introduces three changes that require an explict upgrade step when upgrading from a earlier VOLTTRON version
+VOLTTRON 8 introduces three changes that require an explict upgrade step when upgrading from an earlier VOLTTRON version
 
     1. Dynamic RPC authorization feature - This requires a modification to the auth file. If you have a pre-existing
        instance of VOLTTRON running on an older version, the auth file will need to be updated.
--- a/docs/source/agent-framework/web-framework.rst
+++ b/docs/source/agent-framework/web-framework.rst
@@ -8,6 +8,10 @@ This document describes the interaction between web enabled agents and the Platf
 
 The web framework enables agent developers to expose JSON, static, and websocket endpoints.
 
+.. note::
+    The Platform Web Service Agent also provides a RESTful API which can be used to interact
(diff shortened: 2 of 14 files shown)


craig8:
https://raw.githubusercontent.com/davidraker/volttron/web-api_documentation/docs/source/platform-features/security/volttron-security.rst

I believe that this at the bottom of the page will allow the expansion on the left

```
.. toctree::

   key-stores
   known-hosts-file
   running-agent-as-user

```

davidraker:
I did do this in the platform-features/web-api/introduction.rst file, which is the one linked by the index.rst. It doesn't seem to be working for me, though.

craig8:
after you did this, did you regenerate and then open in a cache free environment.  Are you using an actual server or just viewing the html....

like python -m http.server ... to serve the build/html directory.

davidraker:
I was just visiting the file in a browser after regenerating the html. I thought I had force reloaded, but apparently I did not. I am getting a single sub entry. I should be able to debug why the second entry isn't showing up from here. Should this continue to expand down to nested levels, though, or stop at the second level down?

craig8:
I think you should look at it from a user of the api and do what is best for them....I am not sure which way would be more beneficial.

craig8:
After checking I think we should rename VUI to something more meaningful.
image

In addition, there needs to be some sort of link from the Web Framework down to the RESTful web interface.  This is perhaps outside the initial scope of this, however I think the link would be a good idea in this pr.

craig8:
@davidraker is this ready for a review again?

davidraker:
> @davidraker is this ready for a review again?

I believe it is now. I've added my last changes. I let it get stale, so I could be forgetting something, but after looking it over a bit, I haven't found anything missing. Let me know if you see anything obvious I've forgotten.

craig8:
@davidraker when clicking on the authentication link on the RESTful web api page it doesn't seem to map into the tree.

Issue #2997, opened 2022-09-07
pika import error during volttron startup with zmq only


craig8:
The following error is produced after bootstrapping the develop or releases/8.2 branch.

From the error the auth_protocol should only be imported if rmq is available for the platform.  

(volttron) gridappsd@gridappsd_dev_2004:/repos/volttron-develop$ python bootstrap.py
(volttron) gridappsd@gridappsd_dev_2004:/repos/volttron-develop$ volttron -vv
Traceback (most recent call last):
  File "/repos/volttron-develop/env/bin/volttron", line 11, in <module>
    load_entry_point('volttron', 'console_scripts', 'volttron')()
  File "/repos/volttron-develop/env/lib/python3.8/site-packages/pkg_resources/__init__.py", line 489, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/repos/volttron-develop/env/lib/python3.8/site-packages/pkg_resources/__init__.py", line 2852, in load_entry_point
    return ep.load()
  File "/repos/volttron-develop/env/lib/python3.8/site-packages/pkg_resources/__init__.py", line 2443, in load
    return self.resolve()
  File "/repos/volttron-develop/env/lib/python3.8/site-packages/pkg_resources/__init__.py", line 2449, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/repos/volttron-develop/volttron/platform/main.py", line 73, in <module>
    from volttron.platform.vip.healthservice import HealthService
  File "/repos/volttron-develop/volttron/platform/vip/healthservice.py", line 49, in <module>
    from volttron.utils.rmq_setup import start_rabbit, RabbitMQStartError
  File "/repos/volttron-develop/volttron/utils/rmq_setup.py", line 66, in <module>
    from volttron.platform.auth.auth_protocols.auth_rmq import RMQConnectionAPI
  File "/repos/volttron-develop/volttron/platform/auth/auth_protocols/auth_rmq.py", line 29, in <module>
    class RMQClientParameters(Parameters):
  File "/repos/volttron-develop/volttron/platform/auth/auth_protocols/auth_rmq.py", line 35, in RMQClientParameters
    connection_params: pika.ConnectionParameters = None
NameError: name 'pika' is not defined

schandrika:
HealthService is not even using those RMQ classes. Unused imports can be removed. Will do

PR #3001, opened 2022-09-08
8.2 authfixes

shwethanidd:
# Description

Protecting rabbitmq file imports within guard condition 

Fixes: #2997


craig8 (changes requested):
Just have the one comment about some commented out code but ok otherwise.

PR #3007, opened 2022-09-12
Fixed up VCTL tests

craig8:


volttrontesting/platform/control_tests/test_vctl_commands.py
```diff
@@ -17,47 +17,47 @@
...
     p = subprocess.Popen(
-        ["volttron-ctl", "list"],
+        ["volttron-ctl", "peerlist"],
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
     )
     stdout, stderr = p.communicate()
     try:
-        assert "No installed Agents found" in stderr.decode("utf-8")
+        assert "VOLTTRON is not running." in stderr.decode("utf-8")
```
schandrika, line 60: This is a same assert as in the previous test case. this test case should test that peerlist is returned successfully

schandrika (changes requested):

PR #3009, opened 2022-09-13
Fix web based tests

craig8:


craig8:
Just figured out that the web tests weren't actually testing against rmq!  

https://github.com/VOLTTRON/volttron/blob/releases/8.2/volttrontesting/fixtures/volttron_platform_fixtures.py#L267

The pop removes the param entry and then the call to build_wrapper always has the default data passed!

PR #3019, opened 2022-09-28
8.2 doc updates for non auth and left tree reorg

schandrika:
Also included a minor bug fix to [rmq_config_params.py] from develop


craig8:
Where can I see this on rtd?

schandrika:
https://cs-volttron.readthedocs.io/en/8.2_doc_update/

PR #3038, opened 2022-09-29
Minor fix for volttron-upgrade command

### LINKED ISSUE #3037: volttron-upgrade command fails when no agents are installed in current volttron home
volttron-upgrade command fails when no agents are installed in current volttron home

schandrika:
Fix #3037


craig8:
The proper way to do this is to have the PR be independent of each other...so you should go back to releases/8.2 and then do a branch from there to add this PR rather than from the other PR and extending.

PR #3054, opened 2022-10-20
fix for issue 3053. Added check to determine if variable meta is a dict

### LINKED ISSUE #3053: Use jsonb datatype within postgres/timescale historian
**Is your feature request related to a problem? Please describe.**
I changed my local database to use jsonb for the metadata column in the topics table.  When the agent starts up for the first time everything is fine, however on subsequent starts the agent fails to start due to jsonb returning a dictionary rather than the text for the metadata.

As seen in the following stack trace.

```
2022-10-19 11:51:25,490 (sqlhistorianagent-4.0.0 954) volttron.platform.agent.base_historian ERROR: Failed to setup historian!
Traceback (most recent call last):
  File "/home/volttron/volttron_2222/volttron/platform/agent/base_historian.py", line 1298, in _historian_setup
    self.historian_setup()
  File "/home/volttron/volttron_2222/volttron/utils/docs.py", line 47, in f
    return self.mthd(obj, *args, **kwargs)
  File "/home/volttron/.volttron_2222/agents/06fd34af-72e6-4d25-b944-c18c38f48c06/sqlhistorianagent-4.0.0/sqlhistorian/historian.py", line 352, in historian_setup
    topic_meta_map = self.bg_thread_dbutils.get_topic_meta_map()
  File "/home/volttron/volttron_2222/volttron/platform/dbutils/postgresqlfuncts.py", line 353, in get_topic_meta_map
    meta_map = {tid: jsonapi.loads(meta) if meta else None for tid, meta in rows}
  File "/home/volttron/volttron_2222/volttron/platform/dbutils/postgresqlfuncts.py", line 353, in <dictcomp>
    meta_map = {tid: jsonapi.loads(meta) if meta else None for tid, meta in rows}
  File "/usr/lib/python3.8/json/__init__.py", line 341, in loads
    raise TypeError(f'the JSON object must be str, bytes or bytearray, '
TypeError: the JSON object must be str, bytes or bytearray, not dict
```

**Describe the solution you'd like**
I think it would be easily handled in the current code structure by determining if the passed column is a dictionary or not before attempting to load it.

**Describe alternatives you've considered**
Remove all of the text and go with jsonb for all

ntenney:
# Description

Fixes #3053

## Type of change

# How Has This Been Tested?

This code has been run against databases containing jsonb type and strings for metadata. In addition, I used this code against a local list that simulated the results coming from the database, and also tested it against a situation that should never arise in the wild (strings mixed with objects in the metadata result column). This was tested using the 8.2 deploy of campus.


volttron/platform/dbutils/postgresqlfuncts.py
```diff
@@ -350,7 +350,7 @@ def get_topic_meta_map(self):
             'SELECT topic_id, metadata '
             'FROM {}').format(Identifier(self.meta_table))
         rows = self.select(query)
-        meta_map = {tid: jsonapi.loads(meta) if meta else None for tid, meta in rows}
+        meta_map = {tid: meta if meta and type(meta) == dict else jsonapi.loads(meta) if meta else None for tid, meta in rows}
```
Siecje, line 353: `type(meta) == dict` should be `isinstance(meta, dict)`

craig8: Either works...

```
>>> a = {"alpha": "beta"}
>>> type(a)
<class 'dict'>
>>> type(a) == dict
True
>>> isinstance(a, dict)
True
```

unless I am missing something in the above code?

Siecje: It depends if you want it to work with subclasses of `dict` including `defaultdict` and `OrderedDict`.

```Python
>>> class Foo(dict):
...     pass
... 
>>> type(Foo())
<class '__main__.Foo'>
>>> isinstance(Foo(), dict)
True
>>> from collections import OrderedDict, defaultdict
>>> a = OrderedDict()
>>> b = defaultdict()
>>> type(a)
<class 'collections.OrderedDict'>
>>> type(b)
<class 'collections.defaultdict'>
>>> type(b) == dict
False
>>> isinstance(a, dict)
True
>>> isinstance(b, dict)
True
>>> 
```

craig8: @ntenney as a future proofing please use isinstance rather than type(meta)

volttron/platform/dbutils/postgresqlfuncts.py
```diff
@@ -350,7 +350,7 @@ def get_topic_meta_map(self):
             'SELECT topic_id, metadata '
             'FROM {}').format(Identifier(self.meta_table))
         rows = self.select(query)
-        meta_map = {tid: jsonapi.loads(meta) if meta else None for tid, meta in rows}
+        meta_map = {tid: meta if meta and isinstance(meta, dict) else jsonapi.loads(meta) if meta else None for tid, meta in rows}
```
bonicim, line 353: Just for my understanding, is this nested ternary expression one-liner equivalent to the following for-loop?

```python
meta_map = {}
for tid, meta in rows:
   if meta:
       if meta and isinstance(meta, dict):
           meta_map[tid] = meta
       else:
           meta_map[tid] = jsonapi.loads(meta) 
   else:
       meta_map[tid] = None
```

craig8: Yes

awoz: Just curious. Is obfuscation best practice for this project?

craig8: @awoz No it's not

I think nested if looping this way is harder to read as well.  The original meta_map was the right level of complexity, but I did have a hard time following the new code.

@ntenney can you replace the code with something more user friendly.

bonicim: > @awoz No it's not
> 
> I think nested if looping this way is harder to read as well. The original meta_map was the right level of complexity, but I did have a hard time following the new code.
> 
> @ntenney can you replace the code with something more user friendly.

@ntenney If it helps, feel free to copy and/or modify the `for loop` in my initial comment above ^^

ntenney: I'm not a huge fan of splitting something that takes 1 line into 9. Would this be more palatable?

`meta_map = {tid: meta if meta and type(meta) == dict else (json.loads(meta) if meta else None) for tid, meta in rows}`

schandrika: @ntenney splitting the long line or adding the parenthesis doesn't make it any clearer for me. I do get what the code is doing but I agree that the broken down for loop was much easier to read.  I might have voted for the compact format if there was just a single if inside but the second if inside for loop inside the {} really pushes it over the edge for me

awoz: It is my general sense that all source code should be written "for other" developers who may not be familiar with your code. For them to be successful in their roles, source code hints in the form of comments or clean format are essential. This principle allows for an efficient coding, debug and review cycle, with a timely delivery of solutions. I'm pretty sure that understanding and debugging / single stepping a one-liner is not very palatable.

craig8:
@ntenney Thank you!

PR #3057, opened 2022-11-10
Add skip test logic when pika not installed

bonicim:
# Description

Some tests require RMQ to be installed which in turn requires pika. When running these tests, `pytest` will mark those tests as 'ERROR'. If a volttron platform does not install RMQ, then running tests should not result in 'ERROR'; instead, those tests should be skipped following the example in this [previous PR](https://github.com/VOLTTRON/volttron/pull/3003).

## Type of change

# How Has This Been Tested?

```
pytest -m vc
pytest -m vcp
pytest -m zmq
pytest -m driver
``` 

**Test Configuration**:
Ubuntu 18.04
Python 3.10

# Checklist:

### DIFF
--- a/volttrontesting/platform/test_basehistorian.py
+++ b/volttrontesting/platform/test_basehistorian.py
@@ -13,7 +13,12 @@ from volttron.platform.vip.agent.results import AsyncResult
 from volttron.platform.vip.agent.subsystems.query import Query
 from volttrontesting.utils.utils import AgentMock
 from time import sleep
-from volttrontesting.platform.test_instance_setup import create_vcfg_vhome
+
+from volttron.platform import is_rabbitmq_available
+if is_rabbitmq_available():
+    from volttrontesting.platform.test_instance_setup import create_vcfg_vhome
+else:
+    pytest.skip("Pika is not installed", allow_module_level=True)
 
 
 class QueryHelper:
--- a/volttrontesting/platform/test_instance_setup.py
+++ b/volttrontesting/platform/test_instance_setup.py
@@ -8,11 +8,17 @@ from volttron.platform import is_rabbitmq_available
 from volttron.platform.instance_setup import _is_agent_installed
 from volttron.utils import get_hostname
 from volttron.platform.agent.utils import is_volttron_running
-from volttrontesting.fixtures.rmq_test_setup import create_rmq_volttron_setup
+from volttron.platform import is_rabbitmq_available
+
+RMQ_TIMEOUT = 600
+HAS_RMQ = is_rabbitmq_available()
+if HAS_RMQ:
+    from volttrontesting.fixtures.rmq_test_setup import create_rmq_volttron_setup
+else:
+    pytest.skip("Pika is not installed", allow_module_level=True)
+
 from volttrontesting.utils.platformwrapper import create_volttron_home
 
-HAS_RMQ = is_rabbitmq_available()
-RMQ_TIMEOUT = 600
 
 '''
 Example variables to be used during each of the tests, depending on the prompts that will be asked
--- a/volttrontesting/services/historian/test_multiplatform.py
+++ b/volttrontesting/services/historian/test_multiplatform.py
@@ -57,7 +57,13 @@ from volttrontesting.utils.utils import get_rand_vip, get_hostname_and_random_po
 from volttrontesting.utils.platformwrapper import PlatformWrapper
 from volttrontesting.fixtures.volttron_platform_fixtures import get_rand_vip, \
     get_rand_ip_and_port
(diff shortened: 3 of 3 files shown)


volttrontesting/platform/test_basehistorian.py
```diff
@@ -13,7 +13,12 @@
 from volttron.platform.vip.agent.subsystems.query import Query
 from volttrontesting.utils.utils import AgentMock
 from time import sleep
-from volttrontesting.platform.test_instance_setup import create_vcfg_vhome
+
+from volttron.platform import is_rabbitmq_available
+if is_rabbitmq_available():
+    from volttrontesting.platform.test_instance_setup import create_vcfg_vhome
+else:
+    pytest.skip("Pika is not installed", allow_module_level=True)
```
schandrika, line 21: the create_vcfg_vhome method seem to be create a volttron home and nothing else. (The method should be name create_vhome )So I don't see why it should depend on pika install

volttrontesting/platform/test_instance_setup.py
```diff
@@ -8,11 +8,17 @@
...
-from volttrontesting.fixtures.rmq_test_setup import create_rmq_volttron_setup
-from volttrontesting.utils.platformwrapper import create_volttron_home
+from volttron.platform import is_rabbitmq_available
 
-HAS_RMQ = is_rabbitmq_available()
 RMQ_TIMEOUT = 600
+HAS_RMQ = is_rabbitmq_available()
+if HAS_RMQ:
+    from volttrontesting.fixtures.rmq_test_setup import create_rmq_volttron_setup
+else:
```
schandrika, line 17: The if part makes sense but in the else why should the entire module be skipped. The module has both rmq and zmq tests.

volttrontesting/services/historian/test_multiplatform.py
```diff
@@ -57,7 +57,13 @@
 from volttrontesting.utils.platformwrapper import PlatformWrapper
 from volttrontesting.fixtures.volttron_platform_fixtures import get_rand_vip, \
     get_rand_ip_and_port
-from volttron.utils.rmq_setup import start_rabbit, stop_rabbit
+
+from volttron.platform import is_rabbitmq_available
+if is_rabbitmq_available():
+    from volttron.utils.rmq_setup import start_rabbit, stop_rabbit
+else:
+    pytest.skip("Pika is not installed", allow_module_level=True)
```
schandrika, line 65: Again, only the rmq test cases should be skipped. you can set a variable inside if and use that to skip the rmq test case alone

schandrika (changes requested):

craigpnnl:
Thanks for this, and apologies it sat so long.

The conditional-import half of this has since landed on `develop`
independently: `test_instance_setup.py` and `test_multiplatform.py` now guard
their RabbitMQ imports with `is_rabbitmq_available()`. What is still missing is
the branch you added for the unavailable case, so with `pika` absent those
names are simply never bound and a test that reaches one raises `NameError`
rather than reporting as a skip.

Because of that overlap this no longer merges, and rebasing it would mean
reworking all three files against code that has moved. Rather than ask you to
do that, the remaining gap is tracked in #3271, which credits the approach here
and decides per module whether a module-level skip is the right shape or
whether a narrower guard is better. This PR stays open until that lands.

craigpnnl:
Closing as already fixed on develop. Thank you for this: the problem was real and
the approach here is the one develop ended up taking.

Checked against develop at ee8883899b01913aa67921a85fe298eb619f9d3b on
2026-09-24, in an environment where pika is NOT installed
(`import pika` raises ModuleNotFoundError):

1. `volttrontesting/platform/test_instance_setup.py` already guards the import:
   `HAS_RMQ = is_rabbitmq_available()` followed by
   `if HAS_RMQ: from volttrontesting.fixtures.rmq_test_setup import create_rmq_volttron_setup`.
   This is the same guard this pull request proposed.

2. `volttrontesting/services/historian/test_multiplatform.py` already imports
   `rmq_skipif` from `volttrontesting/skip_if_handlers.py` and applies it, which
   is the shared version of the pattern.

3. `volttrontesting/platform/test_basehistorian.py` has no guard of its own, but
   it no longer needs one: its only rmq-adjacent import is `create_vcfg_vhome`
   from `test_instance_setup`, and that module now imports cleanly without pika
   because of point 1. Verified by importing the module directly with pika
   absent: it imported successfully.

The branch is also 546 commits behind develop and conflicting, so rebasing it
would cost more than it recovers now that the behavior is in place.

If you find a test that still reports ERROR rather than SKIPPED without pika,
please open an issue with the test id and it will be handled directly.

PR #3104, opened 2023-08-17
Revert "Pip fix"

schandrika:
Reverts VOLTTRON/volttron#3094

This doesn't work with python 3.10 because inside the virtual environment created by bootstrap.py the default pip version is the older 22.0.2 and when install is done within this virtual environment pip commands fail with no such option "--config-settings"

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -137,7 +137,7 @@ def update(operation, verbose=None, offline=False, optional_requirements=[], rab
     # option_requirements contains wheel as first entry
 
     # Build option_requirements separately to pass install options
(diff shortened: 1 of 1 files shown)


craig8:
Modular does not support python3.10 anyways only 3.8 on 20.04

craig8:
Modular is written against 20.04 python 3.8 not python 3.10

schandrika:
> Modular is written against 20.04 python 3.8 not python 3.10

It fails for me on Ubuntu 20.04 and python 3.8 as well

PR #3122, opened 2023-10-02
Updates for tests run on Docker before 9.0 release.

davidraker:
# Description

Updates made during testing of 9.0 release on volttron-docker. Changes include:

- Updated version and added requirements entry (openleadr) for OpenADR.
- Fixed relative file path issue in test for OpenADR
- Backported demo-board test and XFailed battery_meter test for ModbusTK.
- Made XFails specific to the relevant parameterizations for control tests.
- Made rabbit imports conditional in tests for multi-platform and instance setup.

## Type of change

# How Has This Been Tested?

All unit and integration tests were run, and these changes were made to fix tests which were not passing.


schandrika (approved):
test_battery_meter.py could skip the test cases if IP is not set (similar to test_driver_demo_board.py). I don't have access to run modbus tk integration tests. Rest of the changes look good and tests pass on my environment

davidraker:
> test_battery_meter.py could skip the test cases if IP is not set (similar to test_driver_demo_board.py). I don't have access to run modbus tk integration tests. Rest of the changes look good and tests pass on my environment

This is not a remote test. The IP used is a localhost one, so that won't actually be the case. I believe this should be left as XFail.

PR #3142, opened 2023-11-13
Updates for 9

craig8:
# Description

Fixes issue with tests
Updates the version to 9.0.1rc0 as this is the first rc for 9.0


volttron/platform/__init__.py
```diff
@@ -35,7 +35,7 @@
 from urllib.parse import urlparse
 
 from ..utils.frozendict import FrozenDict
-__version__ = '8.2'
+__version__ = '9.0.1rc0'
```
shwethanidd, line 38: I thought we had decided it that version would be 9.x

PR #3146, opened 2023-12-04
Cleanup execute_command by using CalledProcessError

Siecje:


craigpnnl (commented):
Review lane: code quality. Reviewed against `develop` at `e2194c1ddecb87681f9f99f18652bb6f6dbbd335`, with this pull request's head at `677c85fc60f3ce0434d0463516e987d1eb86f47b`. Scope: the full diff as it would land on `develop`, plus the affected files read end to end at that base.

The `subprocess.run(check=True)` refactor is faithful: both branches of the
replaced if/else raised the identical `RuntimeError`, and `subprocess.run`
populates `e.stdout` and `e.stderr` from the PIPE captures the same way the
manual returncode check did. The 27 lines of commented-out dead code removed
alongside it are genuinely dead.

Blast radius is wide but well covered indirectly: `execute_command` is called
from the certificate code, the agent installer and the test harness, and over
forty call sites in the control-command tests exercise it, so a regression here
would surface loudly rather than silently.

One note, not blocking: the copy of `execute_command` in `docs/source/conf.py`
has no call site anywhere in the repository, so that half of the change is inert
either way.

craigpnnl (approved):
Approving. The `subprocess.run(check=True)` refactor is faithful: both branches
of the replaced if/else raised the identical `RuntimeError`, and `subprocess.run`
populates `e.stdout` and `e.stderr` from the PIPE captures the same way the
manual returncode check did.

Blast radius is wide but well covered indirectly: `execute_command` is called
from the certificate code, the agent installer and the test harness, and the
control-command tests exercise it from over forty call sites, so a regression
would surface loudly.

Noted and not blocking: the copy of `execute_command` in `docs/source/conf.py`
has no call site anywhere in the repository, so that half of the change is inert
either way.

CI was retriggered; this branch's only prior run failed during dependency
installation, not in this change.

PR #3153, opened 2024-02-19
Added SSL verification to Home Assistant driver

riley206:
# Description

Updated Home Assistant driver to allow for SSL verification with the option of a self signed local certificate. 

# How Has This Been Tested?

Ran on main, develop and 9.x

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Ubuntu 22.04.3 LTS
* WSL

# Checklist:

### DIFF
--- a/services/core/PlatformDriverAgent/platform_driver/interfaces/home_assistant.py
+++ b/services/core/PlatformDriverAgent/platform_driver/interfaces/home_assistant.py
@@ -5,7 +5,7 @@
 #
 # ===----------------------------------------------------------------------===
 #
-# Copyright 2023 Battelle Memorial Institute
+# Copyright 2024 Battelle Memorial Institute
 #
 # Licensed under the Apache License, Version 2.0 (the "License"); you may not
 # use this file except in compliance with the License. You may obtain a copy
@@ -35,16 +35,20 @@ import requests
 from requests import get
 
 _log = logging.getLogger(__name__)
-type_mapping = {"string": str,
-                "int": int,
-                "integer": int,
-                "float": float,
-                "bool": bool,
-                "boolean": bool}
+type_mapping = {"string": str, "int": int, "integer": int, "float": float, "bool": bool, "boolean": bool}
 
 
 class HomeAssistantRegister(BaseRegister):
-    def __init__(self, read_only, pointName, units, reg_type, attributes, entity_id, entity_point, default_value=None,
+
+    def __init__(self,
+                 read_only,
+                 pointName,
+                 units,
+                 reg_type,
+                 attributes,
+                 entity_id,
+                 entity_point,
+                 default_value=None,
                  description=''):
         super(HomeAssistantRegister, self).__init__("byte", read_only, pointName, units, description='')
         self.reg_type = reg_type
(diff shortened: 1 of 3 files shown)


services/core/PlatformDriverAgent/platform_driver/interfaces/home_assistant.py
```diff
@@ -57,7 +61,7 @@ def __init__(self, read_only, pointName, units, reg_type, attributes, entity_id,
 def _post_method(url, headers, data, operation_description):
     err = None
     try:
-        response = requests.post(url, headers=headers, json=data)
+        response = requests.post(url, headers=headers, json=data, verify=self.verify_option)
```
craig8, line 64: What is verify_option ...I don't see this anywhere in the code.

services/core/PlatformDriverAgent/platform_driver/interfaces/home_assistant.py
```diff
@@ -72,29 +76,39 @@ def _post_method(url, headers, data, operation_description):
...
+            raise ValueError("URL is required.")
         if self.access_token is None:
             _log.error("Access token is not set.")
             raise ValueError("Access token is required.")
-        if self.port is None:
-            _log.error("Port is not set.")
-            raise ValueError("Port is required.")
+
+        if not self.verify_ssl:
+            import urllib3
```
craig8, line 103: Why not just do this at the top of the module regardless of whether we use ssl or not?

craig8 (changes requested):

services/core/PlatformDriverAgent/platform_driver/interfaces/home_assistant.py
riley206: its on line 107. This is for the requests library, verify_option will eaither be a path to a crt file or will be False and will not use SSL.

services/core/PlatformDriverAgent/platform_driver/interfaces/home_assistant.py
riley206: its on line 107. This is for the requests library, verify_option will eaither be a path to a crt file or will be False and will not use SSL.

services/core/PlatformDriverAgent/platform_driver/interfaces/home_assistant.py
riley206: I was just thinking that it could save on some resources but its probably insignificant.

craigpnnl:
Thank you for the SSL-verification work on the Home Assistant driver; the
feature is worth having, but this patch cannot be carried as written.

The new `verify=self.verify_option` argument is added inside `_post_method`,
which is a module-level function, not a method, so it has no `self` in scope:
calling it raises `NameError: name 'self' is not defined` on every write (light,
thermostat, switch and the rest). On the current tree `_post_method` is defined
at column 0 between `HomeAssistantRegister` and `Interface`, which is what makes
the reference unresolvable.

The branch also targets `releases/9.x`, which has not moved in almost two years;
our active line is `develop`.

Closing this one, and we would welcome a fresh pull request against `develop`
with the scoping issue fixed: the verify option needs to reach `_post_method`
as an argument, or the function needs to become a method of `Interface`.

PR #3166, opened 2024-04-02
Improve platform.messaging imports

Siecje:


craigpnnl (commented):
Review lane: code quality. Reviewed against `develop` at `e2194c1ddecb87681f9f99f18652bb6f6dbbd335`, with this pull request's head at `3ea14964a407f028523ae82f52344389ac3f9ca1`. Scope: the full diff as it would land on `develop`, plus the affected files read end to end at that base.

The star-import to explicit-import change preserves the re-exported names. The
only external attribute-style caller is `messaging.Socket()`, used in
`volttron/platform/agent/base.py`, and it is unchanged. The dropped
`parse_timestamp_string` import in `health.py` is confirmed unused: it has
exactly one occurrence in the file, its own import line.

No test is added, which is correct for an import-only change; nothing behavioral
happens here that a test could assert.

craigpnnl (approved):
Approving. Verified against `develop`: the star-import to explicit-import change
preserves the re-exported names, and `messaging.Socket()` (the only external
attribute-style caller, in `volttron/platform/agent/base.py`) is unchanged. The
dropped `parse_timestamp_string` in `health.py` has exactly one occurrence in
the file, its own import line.

No test accompanies it, which is right for an import-only change: nothing
behavioral happens here that a test could assert.

CI was retriggered because this branch's only runs predated more than a year of
`develop`.

PR #3167, opened 2024-04-02
Remove unused imports

Siecje:


shwethanidd (approved):

Siecje:
Can this be merged?

craigpnnl (commented):
Review lane: code quality. Reviewed against `develop` at `e2194c1ddecb87681f9f99f18652bb6f6dbbd335`, with this pull request's head at `09950d9876db15b07fd8a4b789baf073abf22db2`. Scope: the full diff as it would land on `develop`, plus the affected files read end to end at that base.

All four removals are confirmed dead. Each removed name was checked against the
whole file it was removed from and has exactly one occurrence, the import line
itself, so nothing can notice the removal.

Still applies cleanly to `develop` despite its age.

craigpnnl:
Checked one thing before merging that a plain unused-import review would not
cover, and it comes out clean.

Two of the removals in `services/core/IEEE_2030_5/ieee_2030_5/agent.py` sit
inside a `try/except ImportError` compatibility shim. Narrowing the imports in a
`try` changes what can raise, so in principle it could change which branch is
taken, independently of whether the names are used.

It cannot here. `volttron.client` does not exist in this repository, and neither
it nor `volttron.utils.commands` is importable in the test environment:

```
FAILS  volttron.client                                -> ModuleNotFoundError
FAILS  volttron.client.vip.agent                      -> ModuleNotFoundError
FAILS  volttron.client.vip.agent.subsystems.query     -> ModuleNotFoundError
FAILS  volttron.utils.commands                        -> ModuleNotFoundError
```

The `try` branch therefore always falls through to `except` regardless of this
change, so the modular-side edit is inert in this distribution. The `except`
branch is the one that runs, and every name removed from it is genuinely unused:
`RPC`, `Core`, `PubSub` and `Query` appear only on the four import lines
themselves (two per branch) and nowhere else in the file; `STATUS_GOOD` and
`pformat` appear once each.

One note for anyone reading the counts later: a word-boundary grep for `field`
returns three lines, which looks like a live use. Two of them are the English
word "field" in a docstring and a comment, at lines 184 and 368. It is unused as
a symbol. Control for that check: `Agent` returns 4 lines and `dataclass` 2, so
the count can exceed 1 when a name really is used.

Retriggering CI, which has not run on this branch at all. Note that no workflow
in this repository collects `services/core/IEEE_2030_5/` or
`services/core/DNP3OutstationAgent/`, so a green run will not exercise these
files; the checks above are the actual evidence, not the CI result.

Issue #3168, opened 2024-04-12
@RPC.allow restrictions are not applied after restart of VOLTTRON


schandrika:
**Describe the bug**
Auth restrictions on RPC methods specified using @RPC.allow  decorators are not enforced on existing agents after a restart of VOLTTRON server. These are enforced on a newly installed agent or after auth update using "vctl auth" commands. 

**To Reproduce**
1. Start VOLTTRON
2. Install agent, say AgentA, with RPC exported method with a auth restriction using the @RPC.allow decorator
3. Install a second agent, say CallerAgent, and call the auth restricted method of AgentA on onstart method
4. Start AgentA and CallerAgent
5. This should cause the RPC call to fail with a error clearly mentioning that CallerAgent does not have the necessary capability to access AgentA's method
6. Restart VOLTTRON
7. Restart AgentA
8. Restart CallerAgent
9. Observe there is no auth error and CallerAgent is able to call the auth restricted method of AgentA

**Expected behavior**
Auth rules should be enforced for newly installed agents and existing agents

PR #3169, opened 2024-04-12
Fix for security issue #3168

schandrika:
Fix and test case for issue #3168

### DIFF
--- a/volttron/platform/main.py
+++ b/volttron/platform/main.py
@@ -435,9 +435,6 @@ class Router(BaseRouter):
     #    return result
 
     def handle_subsystem(self, frames, user_id):
-        # _log.debug(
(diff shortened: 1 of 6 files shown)


craig8:
@schandrika looks like there are a couple of failures.  The utils looks like a AMQP timeout.  Like the rmq server stopped quicker than we would like.  

However, the auth seems to be a genuine error.  https://github.com/VOLTTRON/volttron/actions/runs/8665217491/job/23763385735#step:4:2167  
Please take a look and run this on your system.

schandrika:
@craig8 The auth test pass for me locally. The one failing test in auth here, does depend on gevent.sleep(4) so could be a timing issue. 
Testutils failure is also a final cleanup timeout and the tests themselves pass

craig8 (approved):

craig8 (approved):

Issue #3173, opened 2024-04-20
Changes to argparse break handling of single dash arguments in platform/config.py.


davidraker:
Recent changes to the argparse module in cpython will break the ability of volttron to parse single dash arguments (e.g., "-vv"). This prevents the platform from being started with those options.  The breaking change, which will affect python >=3.11.9 and >=3.12.3, occured here:

https://github.com/python/cpython/pull/115675

The change is to the return signature of ArgumentParser._parse_optional(). It previously returned a 3-tuple: (action, option_string, explicit_arg). It now returns a 4-tuple: (action, option_string, sep, explicit_arg).

**To Reproduce**
Start VOLTTRON on 3.12:
1. unpin gevent, pyzmq, and requests (updating these is already a separate issue).
2. run bootstrap.
3. volttron -vv

You will get a "too many values to unpack" error which kills VOLTTRON early in the startup process.

**Expected behavior**
VOLTTRON starts normally.

Issue #3158, opened 2024-03-25
Python 3.11 + support.


rlutes:
**Describe the bug**
When bootstrapping the platform with version Python 3.11 or greater, the build fails.  

**To Reproduce**
Clone repository and run bootstrap.py.

**Expected behavior**
bootstrapping the platform runs without error.

**Additional context**
by editing the requirements.py file and unpinning version for pyzmq and gevent, the bootstrap process will succeed.

PR #3174, opened 2024-04-20
3.12 updates

davidraker:
# Description

Updates necessary to bootstrap and run VOLTTRON on python versions up to 3.12.

Fixes # (issue)
#3158 #3173
## Type of change

# How Has This Been Tested?
Confirmed that these changes allow VOLTTRON to bootstrap and start on Ubuntu 24.04 LTS (beta).

# Checklist:

### DIFF
--- a/requirements.py
+++ b/requirements.py
@@ -30,12 +30,12 @@
 # wheel version 0.31 has removed metadata.json file
 # https://github.com/pypa/wheel/issues/195
 # so sticking to 0.30 for now. Could upgrade to wheel 0.31 with code changes
-option_requirements = [('pip==24.0', []), ('wheel==0.30', []), ('pyzmq==22.2.1', ['--zmq=bundled'])]
+option_requirements = [('pip==24.0', []), ('wheel==0.30', []), ('pyzmq==26.0.2', ['--zmq=bundled'])]
 
 
-install_requires = ['gevent==21.12.0',
-                    'grequests==0.6.0',
-                    'requests==2.23.0',
+install_requires = ['gevent==24.2.1',
+                    'grequests==0.7.0',
+                    'requests==2.31.0',
                     'idna<3,>=2.5',
                     'ply==3.11',
                     'psutil==5.9.1',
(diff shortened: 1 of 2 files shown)


craig8:
@davidraker please confirm that these updates are backward compatible with 3.8 and 3.10.  I.e one of the requirements hasn't dropped support for them.  

If so we may need to discuss further.  Also, please mitigate the test run errors that have shown up.

davidraker:
> @davidraker please confirm that these updates are backward compatible with 3.8 and 3.10. I.e one of the requirements hasn't dropped support for them.
> 
> If so we may need to discuss further. Also, please mitigate the test run errors that have shown up.

@craig8 Will do. It is definitely still a draft. Just starting the conversation with something that tentatively works on 24.04. This about to come up a lot more now.

craigpnnl (commented):
Review lane: code quality. Reviewed against `develop` at `e2194c1ddecb87681f9f99f18652bb6f6dbbd335`, with this pull request's head at `87d8f4afc97c047cfc6f0b2106dc43899d4c8b92`. Scope: the full diff as it would land on `develop`, plus the affected files read end to end at that base.

**Verdict: changes requested.** Two separate problems, one per file.

**`requirements.py` is already done.** `develop` already carries
`pyzmq==26.0.2`, `gevent==24.2.1`, `grequests==0.7.0` and `requests==2.31.0`,
verbatim. A three-way merge drops this hunk silently because both sides have
converged on the same text, so this half contributes nothing.

**`config.py` does not do what it claims, on the versions it targets.** This was
checked by running it rather than by reading it. On Python 3.13,
`ArgumentParser._parse_optional` returns a one-element **list** wrapping a
four-tuple, not a bare four-tuple. The added `len(option_tuple)` check therefore
sees `len == 1`, matches neither the `== 3` branch nor the `== 4` branch, and
falls into this patch's own `else: raise ValueError`.

Run against a real `_parse_optional('-xfoo')` result: Python 3.10 takes the
three-element branch and succeeds, Python 3.13 raises
`ValueError('Unable to parse options: "-xfoo".')`. So on 3.10 the change is a
no-op, and on 3.13 it converts a working path into an exception.

The underlying goal is worth pursuing; this implementation needs to be written
against the actual return shape of `_parse_optional` on each targeted version,
with a test that pins that shape.

PR #3175, opened 2024-04-22
Fixes process overload from file events

### LINKED ISSUE #3172: 8.2 Authfile Still loading constantly
**Describe the bug**
The file watcher is filing continuously rather than on file changed

**To Reproduce**
Start volttron 8.2 and watch log

**Expected behavior**
The auth file is only reloaded on change

**Screenshots**

**Operating System (please complete the following information):**
 - OS: Ubuntu 22.04
 - Volttron Version Releases/8.2

**Additional context**

acedrew:
# Description

Changes file watcher classes to use "on_closed" rather than "on_any_change" event type, reducing frequency of events.
Fixes #3172 

## Type of change

- [*] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [ ] Deployed to platforms that were experiencing extreme resource utilization and verified utilization dropped significantly
- [ ] Deployed to platforms with ongoing system profiling and verified the hotpaths in the platforms changed.

# Checklist:

### DIFF
--- a/volttron/utils/__init__.py
+++ b/volttron/utils/__init__.py
@@ -122,7 +122,7 @@ class VolttronHomeFileReloader(PatternMatchingEventHandler):
         _log.debug("patterns is {}".format([get_home() + '/' + filetowatch]))
         self._callback = callback
 
-    def on_any_event(self, event):
+    def on_closed(self, event):
         _log.debug("Calling callback on event {}. Calling {}".format(event, self._callback))
         try:
             self._callback()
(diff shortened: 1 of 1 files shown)


craig8 (approved):

craig8:
@acedrew This actually breaks the reloading of the web-users.json file.  However I found a fix, remove gevent_watchdog and just use watchdog.  I am testing now to make sure it doesn't have any other concequences.

craig8:
I don't know...my guess is that we are reading auth.json in more than we do the web.  So there are places that read it regardless of it being actually run through the event itself.  Web users is only read/written in a couple of places.

craig8:
I suppose the other thing could be some gevent stuff happening that unless you use all_events then it's not getting propagated correctly?  Not 100% sure

craig8:
@acedrew I think I found the issue.  So the webusers uses a PersistentDict to write the data to a file.  It writes data to a temp file and the copies the file into the proper space.  This apparently is not what the filewatch mechanisms is expecting so it sends an infinite list of opened events!  Sigh!

PR #3178, opened 2024-05-01
Modbustk error handler doesn't include .message

craig8:
# Description

This pull request fixes the error handler that relies on .message being set on Modbustk Exception handler.  The Modbustk exception handler does not set message so this handler never would give an accurate error message.


craig8:
@davidraker  please look at this

davidraker (commented):
This checks out, and looks correct. My only question is why we are detecting the exception type in an if statement instead of adding a separate except blocks for ModbusError and AttributeError? I don't think there is a functional difference, but it would be a little cleaner, I think?

PR #3196, opened 2024-11-11
Pinned setuptools in requirements.py. Newer versions that 70 are inco…

davidraker:
Pinned setuptools in requirements.py. Newer versions that 70 are incompatible with wheel 0.30.0.

# Description

Changed pin on setuptools to be between 40.0.0 and 70.0.0. 

Fixes # (issue)

## Type of change

# How Has This Been Tested?

Installed on new environment and tested installing agents both with and without the pin.  It will fail if setuptools is newer than 70.0.0. 
# Checklist:

### DIFF
--- a/requirements.py
+++ b/requirements.py
@@ -42,7 +42,7 @@ install_requires = ['gevent==21.12.0',
                     'python-dateutil==2.8.2',
                     'pytz==2022.1',
                     'PyYAML==6.0',
-                    'setuptools>=40.0.0',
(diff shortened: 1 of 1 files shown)


craigpnnl:
This targets `releases/9.x`, which has not moved since late 2023; our active
line is `develop`. The change itself is no longer needed there either:
`develop`'s `requirements.py` already pins `setuptools>=40.0.0,<=70.0.0`,
identical to what this pull request proposes.

Thank you for the fix; closing as already applied.

PR #3220, opened 2025-10-29
fix: handle option tuple unpacking for Python versions >= 3.12

bonicim:
# Description

Fix tuple unpacking when running Volttron on Python 3.12>= while supporting backwards compatibility on older versions of Python. Duplicates work done in modular Volttron. See https://github.com/eclipse-volttron/volttron-core/pull/268

## Type of change

# How Has This Been Tested?

**Test Configuration**:
* OS: Ubuntu 24.03  
* Python: 3.12
* Hardware: Virtual Box 

Installed and ran Volttron. Successfully started and stopped Volttron; also installed Listener, Platform Driver, and Historian agents.  

# Checklist:


bonicim:
This pull request fixes the parsing issue already addressed in Modular Voltton. See https://github.com/eclipse-volttron/volttron-core/pull/268

craigpnnl:
Thank you for taking on Python 3.12+ support here; it is a real gap, and it is
still open.

This specific patch does not close it, though. On Python 3.13, `argparse`'s
`_parse_optional` returns the option as a one-element **list** wrapping the
tuple rather than a bare tuple, so the `option_tuple` unpacked here is that
list, not the 4-tuple the new branch expects. Running this branch's own logic
against a real `_parse_optional('-xfoo')` result raises
`ValueError: not enough values to unpack (expected 4, got 1)`. On 3.10 the new
branch is never taken, so the change is a no-op there.

It also targets `main` rather than our active `develop` branch, where the
version bumps in `requirements.py` have moved past what this proposes: applying
it as-is would downgrade the docker pin from 7.2.0 back to 7.1.0 and revert a
recent fix.

Closing this one. The underlying gap is tracked separately, and a correct fix
written against the actual return shape of `_parse_optional`, with a test
pinning that shape, would be very welcome against `develop`.

PR #3228, opened 2026-05-07
Weather Cache Fixes

davidraker:
This pull request significantly improves the cache management logic in the `base_weather.py` module, focusing on more accurate cache size calculation, safer deletion of old records, and better error handling. The changes ensure that the cache does not exceed its maximum allowed size and prevent infinite loops during deletion attempts.

**Cache management improvements:**

* Updated the `page_count` method to use a more accurate SQL query that calculates used pages by subtracting freelist pages, providing a true representation of used cache space.
* Enhanced the `manage_cache_size` method to prevent infinite loops by introducing maximum attempt limits for both targeted and general deletion operations.
* Improved the deletion logic to commit changes after each deletion attempt, ensuring that the page count reflects the latest state of the database.
* Added warnings and error logging if the cache cannot be reduced below the maximum size or if deletion attempts are exhausted, making issues easier to diagnose.

**Code robustness and clarity:**

* Added error handling around the cache management logic to log exceptions and re-raise them, increasing reliability and debuggability.
* Switched to using `get_aware_utc_now()` for time calculations, improving time zone handling. (# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- Installed the updated agent.
- Set to run with a short poll interval.
- Set the max cache limit very small.
- Observed that the cache was being correctly deleted and that the agent no longer became unresponsive when the cache reached its limit.

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain:
* SDK:

# Checklist:

### DIFF
--- a/volttron/platform/agent/base_weather.py
+++ b/volttron/platform/agent/base_weather.py
@@ -1174,6 +1174,8 @@ class BaseWeatherAgent(Agent):
                                            "Weather agent cache is full")
                 status = Status.from_json(self.vip.health.get_status_json())
                 self.vip.health.send_alert(CACHE_FULL, status)
+            else:
+                self.vip.health.set_status(STATUS_GOOD)
         except Exception as error:
             err_msg = "Weather agent failed to write to cache"
             _log.error("{}. Exception:{}".format(err_msg, error))
--- a/volttron/platform/agent/base_weather.py
+++ b/volttron/platform/agent/base_weather.py
@@ -938,7 +938,7 @@ class BaseWeatherAgent(Agent):
             if not generation_time:
                 # in case api does not return details on when this
                 # forecast data was generated
-                generation_time = datetime.datetime.utcnow()
+                generation_time = get_aware_utc_now()
             else:
                 generation_time, oldtz = process_timestamp(
                     generation_time)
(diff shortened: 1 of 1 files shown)


craigpnnl (commented):
Review lane: code quality. Reviewed against `develop` at `e2194c1ddecb87681f9f99f18652bb6f6dbbd335`, with this pull request's head at `a3ea57cb01779f16f17657133c57e087004dc468`. Scope: the full diff as it would land on `develop`, plus the affected files read end to end at that base.

**Verdict: sound fix, one gap.**

The bug is real. `PRAGMA page_count` returns pages ever allocated, which does
not shrink after a `DELETE` without `auto_vacuum` or an explicit `VACUUM`, so
the previous `while page_count >= max_pages` loop could spin without making
progress. Subtracting `freelist_count` tracks used space, which does shrink, and
the attempt caps bound the loop.

The imports the new call sites need are already present at the top of the file,
so the change is self-contained.

**The gap:** the only coverage for this path is an integration test that needs a
live agent and network access, so nothing fast will catch a regression here. A
focused unit test over the page-count arithmetic and the attempt cap, asserting
the computed value rather than absence of a crash, would make this landable
without reservation.

PR #3229, opened 2026-07-01
Handle error num passed as a string in platform.vip.agent.errors.

davidraker:
This pull request makes a small update to the error handling logic in `volttron/platform/vip/agent/errors.py`. The change improves flexibility by allowing the `from_errno` method to accept error numbers as strings prefixed with `'errno.'`, in addition to numeric values.

* Enhanced `from_errno` to accept error numbers as strings like `'errno.EHOSTUNREACH'`, converting them to the corresponding `errno` constant if necessary.

### DIFF
--- a/volttron/platform/vip/agent/errors.py
+++ b/volttron/platform/vip/agent/errors.py
@@ -46,6 +46,8 @@ class VIPError(Exception):
 
     @classmethod
     def from_errno(cls, errnum, msg, *args):
+        if isinstance(errnum, str) and errnum.lower().startswith('errno.'):
+            errnum = getattr(errno, errnum[len('errno.'):])
         errnum = int(errnum)
         return {
             errno.EHOSTUNREACH: Unreachable,


craigpnnl (commented):
Review lane: code quality. Reviewed against `develop` at `e2194c1ddecb87681f9f99f18652bb6f6dbbd335`, with this pull request's head at `8ccff5e1aaaba52f5030d7e0b91ecd2b310d1ba7`. Scope: the full diff as it would land on `develop`, plus the affected files read end to end at that base.

**Verdict: safe, but untested.**

`errno` is already imported, and the `isinstance` guard changes behavior only
for a string input beginning with `errno.`. The sole current caller passes the
value straight off the wire and is always an integer, so nothing existing is
affected either way.

That is also the reservation: no test accompanies this, and no caller in the
repository exercises the new path, so the added branch ships unexercised. A
short unit test calling `from_errno` with both an int and an `errno.` string,
asserting the resulting error type and code, would cover it in a few lines.

craigpnnl:
This fixes more than it appears to, and there is now a concrete reproduction
for it.

The earlier review note on this pull request said the change is safe but that
no caller in the repository exercises the new path. That is true of the
repository's own call sites, and it is not true of the wire. The platform
itself passes the string form.

Reproduced against `develop` e2194c1ddecb87681f9f99f18652bb6f6dbbd335:

```
File "volttron/platform/vip/agent/core.py", line 585, in handle_error
    error = VIPError.from_errno(*args)
File "volttron/platform/vip/agent/errors.py", line 49, in from_errno
    errnum = int(errnum)
ValueError: invalid literal for int() with base 10: 'Errno.EHOSTUNREACH'
<Greenlet: vip_loop> failed with ValueError
...
    raise RuntimeError('VIP loop ended prematurely')
```

`handle_error` receives `'Errno.EHOSTUNREACH'` as a string, `int()` raises, and
the agent's `vip_loop` greenlet dies. The agent never recovers.

In the historian test suite this is not cosmetic. A module-scoped agent is
shared across a whole parametrize group, so once its core dies every later test
in that group fails, and the group's teardown then blocks forever trying to
stop an already-dead core. Two separate issues were filed from the symptoms
before the cause was known, #3273 for the cascade and #3274 for the hang.

Applying exactly the guard proposed here and rerunning the reproducing
invocation:

- without the guard: the second test fails, and teardown hangs (one run left
  alive past 811 seconds with no further output)
- with the guard: `12 passed, 13 warnings in 120.69s`, finishing cleanly with
  no hang

So this two-line change removes a live test cascade and a hang, not just a
theoretical unpacking problem.

Two notes toward landing it:

1. The reproduction above is the regression test this was missing. A unit test
   calling `from_errno` with both an int and an `'Errno.EHOSTUNREACH'` string,
   asserting the resulting error type and code rather than merely that it does
   not raise, would cover it in a few lines and does not need a platform.
2. `getattr(errno, errnum[len('errno.'):])` will raise `AttributeError` on an
   unknown name rather than returning something sensible. Worth deciding
   deliberately whether an unrecognised errno name should raise, or fall back
   to the base `VIPError`, since this code runs on the error path of a message
   loop where an exception is what kills the greenlet in the first place.

craigpnnl:
Retriggering CI on this pull request. Its only runs are from 2026-07-01 and
they fail during dependency installation, not in this change:

```
Building wheel for modbus-tk (setup.py): finished with status 'error'
ModuleNotFoundError: No module named 'wheel.wheelfile'
ERROR: Failed building wheel for modbus-tk
```

That signal is stale. `develop` carries the same `wheel==0.30`,
`modbus-tk==1.1.2` and `treelib==1.6.1` pins today and all eight workflows pass
on it, so the July failure is not reproducible against the current base and
says nothing about a two-line change in `errors.py`.

Closing and reopening to get a run against current `develop`. No change to the
branch or its contents.

craigpnnl (approved):
Approving. The two-line guard is correct and its effect is proven by experiment, not by reading.

The string form reaching `from_errno` kills the agent's `vip_loop` greenlet via `ValueError` at `errors.py:49`. In the historian suite that surfaced as a 40-test cascade (#3273) and an infinite teardown hang (#3274), because a module-scoped agent is shared across a parametrize group: once its core dies every later test talks to a dead agent, and teardown then blocks forever trying to stop it.

Measured on the reproducing twelve-test invocation:

- without this change: the test fails, and teardown hung past 811 seconds before being killed
- with this change: `12 passed, 13 warnings in 120.69s`, finishing cleanly

CI on the rebased head is green where it was red: all nine build jobs failed on 2026-07-01 with a `wheel.wheelfile` dependency error, and none of them has recurred against current `develop`.

The regression test this was missing is #3276, which is RED without this change and green with it. It should land after this one.

One item deliberately left for you rather than changed here: `getattr(errno, errnum[len('errno.'):])` raises `AttributeError` on an unrecognised errno name, confirmed by running it. That is the same unhandled-exception failure mode on the same message-loop error path this change fixes, so it is worth a deliberate decision, but it does not block landing the case we actually hit.

PR #3232, opened 2026-07-10
fix: use platform.node() for AGENT_PING to avoid gevent import deadlock

### LINKED ISSUE #3231: AGENT_PING topic construction deadlocks under gevent on modern CPython (platform.uname subprocess)
volttron/platform/messaging/topics.py builds the AGENT_PING topic at module import time using platform.uname()[1] to obtain the hostname:

```python
AGENT_PING = _("agent/ping/{}/{}/{{cookie}}".format(platform.uname()[1], os.getpid()))
```

platform.uname() does more than read the hostname: it shells out to a subprocess to populate an unrelated "processor" field. VOLTTRON monkey-patches os with gevent (via volttron/utils/__init__.py monkey_patch, fired from vip/__init__.py), and that subprocess call deadlocks under gevent on modern CPython. It reproduces on Python 3.13: the import chain hangs indefinitely, and faulthandler pins the hang at topics.py:91.

Because the auth import chain pulls in topics.py, `import volttron.platform.auth` hangs and the platform auth tests cannot run.

## Impact

Any importer of a module whose import chain reaches messaging/topics.py hangs at import time on affected interpreters. The auth package is the observed case.

## Fix

Use platform.node() instead of platform.uname()[1]. Both return the identical node name, but node() reads the hostname directly with no subprocess, so it is a behavior-preserving change. Verified green on Python 3.10 (48 passed, 70 skipped, 2 xfailed). A regression test asserts that the auth package imports and that AGENT_PING renders real field values (hostname and pid).

## Reproduction

On Python 3.13, `import volttron.platform.auth` hangs; faulthandler shows the stack parked at topics.py:91 inside the platform.uname() subprocess call.

craigpnnl:
Fixes #3231

## Problem

volttron/platform/messaging/topics.py builds the AGENT_PING topic at module import time from platform.uname()[1]. platform.uname() shells out to a subprocess to populate its "processor" field, and under gevent's monkey-patched os that subprocess call deadlocks at import time on modern CPython (observed hanging on Python 3.13). Because the auth import chain reaches topics.py, `import volttron.platform.auth` hangs and the platform auth tests cannot run. faulthandler pins the hang at topics.py:91.

## Fix

Replace platform.uname()[1] with platform.node(). Both return the identical node name, but node() reads the hostname directly with no subprocess. This is a behavior-preserving change that lets the auth package import on modern interpreters while leaving the pinned 3.10 toolchain unaffected.

A regression test (volttrontesting/platform/auth_tests/test_auth_import.py) asserts two things: the auth package imports cleanly, and AGENT_PING renders agent/ping/<node>/<pid>/<cookie> with the real hostname and pid values, not merely that rendering does not crash.

## Test plan

Ran the auth test suite the CI way on Python 3.10:

```
pytest volttrontesting/platform/auth_tests -m auth
```

Result: 48 passed, 70 skipped, 2 xfailed.

The regression test was verified RED on Python 3.13 with the prior platform.uname()[1] code (import deadlocked at roughly 18s) and GREEN with the platform.node() fix.

### DIFF
--- a/volttron/platform/messaging/topics.py
+++ b/volttron/platform/messaging/topics.py
@@ -88,7 +88,12 @@ MARKET_ERROR = _(MARKET_BASE.replace('{subtopic}', 'error'))
 MARKET_RECORD = _(RECORD.replace('{subtopic}', MARKET_CLEAR))
 
 AGENT_SHUTDOWN = _('agent/{agent}/shutdown')
-AGENT_PING = _('agent/ping/{}/{}/{{cookie}}'.format(platform.uname()[1], os.getpid()))
+# Use platform.node() rather than platform.uname()[1] for the hostname. Both
+# return the same node name, but uname() additionally shells out to a
+# subprocess to fill the "processor" field, and that subprocess call deadlocks
+# at import time under gevent's monkey-patched os on newer CPython (observed
+# hanging on 3.13). node() reads the hostname directly with no subprocess.
+AGENT_PING = _('agent/ping/{}/{}/{{cookie}}'.format(platform.node(), os.getpid()))
 
 LOGGER_BASE =_('datalogger')
 LOGGER = _('datalogger/{subtopic}')
--- a/volttrontesting/platform/auth_tests/test_auth_import.py
+++ b/volttrontesting/platform/auth_tests/test_auth_import.py
@@ -1,0 +1,78 @@
+# -*- coding: utf-8 -*- {{{
+# ===----------------------------------------------------------------------===
+#
+#                 Component of Eclipse VOLTTRON
+#
+# ===----------------------------------------------------------------------===
+#
+# Copyright 2023 Battelle Memorial Institute
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy
+# of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


craigpnnl (commented):
Review lane: code quality. Reviewed against `develop` at `e2194c1ddecb87681f9f99f18652bb6f6dbbd335`, with this pull request's head at `cec068db75434ddb48da3f75611b8d28bc8547e9`. Scope: the full diff as it would land on `develop`, plus the affected files read end to end at that base.

**Verdict: no blocking findings.** The claim in the description was checked by
running it, and it holds, though for a more specific reason than stated.

`platform.node()` is `return uname().node` on both 3.10 and 3.13, so the naive
reading is that both call `uname()` and nothing changes. That reading is wrong.
The subprocess enters through `platform.uname_result.processor`, a
`cached_property` backed by `uname -p`. That class defines
`__iter__` as `itertools.chain(super().__iter__(), (self.processor,))` and
`__getitem__` as `tuple(self)[key]`, so **any** index access forces the
`processor` property and its subprocess, whatever index is asked for. Attribute
access never touches it.

Measured directly on 3.10 with `subprocess.check_output` instrumented:
`platform.uname()[1]` made one subprocess call, `platform.node()` called
immediately afterwards made zero, both returning the identical hostname.

This means the change is not only a newer-interpreter accommodation: it removes
a subprocess call from module import on every version, 3.10 included.

The new test asserts the rendered topic's segments equal the real hostname, pid
and cookie values, which is a field-value assertion rather than a non-crash
check, and it needs no live platform instance. Good test.

Issue #3238, opened 2026-09-17
Test harness dynamic_agent lacks the control capabilities ControlService now requires


craigpnnl:
**Problem.** `PlatformWrapper` uses `dynamic_agent` to call `control.remove_agent` (`remove_all_agents`) and `control.shutdown` (`stop_platform`, `shutdown_platform`), and tests call `prioritize_agent` through it. Since b83b8ea6b these methods require `install_remove_agents` and `start_stop_agents`.

**Current grants.** The harness pre-seed gives `dynamic_agent` only `edit_config_store` and `allow_auth_modifications`. When pre-seeding is skipped, `build_agent` grants only `edit_config_store` for the agent's own identity.

**Effect.** 31 failures on auth-enabled parameters, plus failures that look like isolation problems, because `remove_all_agents` no longer cleans up between tests.

**Acceptance criteria:**
1. **Entry contents.** On an auth-enabled instance, the auth.json entry for `dynamic_agent` holds exactly:
   * `edit_config_store` with `{"identity": "/.*/"}`
   * `allow_auth_modifications` with `None`
   * `clear_agent_status`, `install_remove_agents`, `start_stop_agents`, `stop_platform` and `tag_agents`, each with `None`, named through the constants in `volttron.platform.agent.known_identities`

   A test reads the file through `AuthFile` and asserts every key and value.
2. **Other startup paths.** Criterion 1 also holds when auth.json already had allow entries before `startup_platform`, and when an existing `dynamic_agent` entry lacked the control capabilities. Both have tests.
3. **Default build_agent grant unchanged.** An agent created by `build_agent()` without a `capabilities` argument still has exactly `{"edit_config_store": {"identity": <its identity>}}`. A test asserts it.
4. **Cleanup works.** New test: install ListenerAgent, call `remove_all_agents()`, and `list_agents` returns `[]`. This test fails at d68dff037 with `requires capabilities {'install_remove_agents'}`.
5. **No driver capabilities.** No harness identity receives `driver_write` or `driver_override`.
6. **Refusal tests unchanged.** Outcomes of volttrontesting/platform/auth_tests and tests/unit/test_rpc_allow_control.py do not change.
7. **Newly passing (37 ids).** Any of these still failing is reported on this issue with its message.
[...]

Issue #3248, opened 2026-09-22
Test harness pre-seed keeps only the last auth entry it adds


craigpnnl:
## Summary

The test harness's startup pre-seed constructs one `AuthFile` object and calls `add()` on it repeatedly, in `PlatformWrapper.startup_platform` (`volttrontesting/utils/platformwrapper.py`, `authfile = AuthFile()` followed by several `authfile.add(entry)` calls, around lines 813 to 851 at `develop` d68dff037b680383442576c634980a18a6aed8b7 plus the branch for VOLTTRON/volttron#3238).

`AuthFile` writes from the entry snapshot it loaded when it was constructed, and the write path does not reload that snapshot, so each `add()` writes the original set plus its own entry. The last write wins, and the entries added before it are not on disk.

Reported during review of #3238, measured by replicating that block verbatim: 1 entry on disk where 9 were added. Control: a fresh `AuthFile` per `add()` kept every entry in the same probe. This is a reviewer's measurement, from a replication of the block rather than from a platform run.

This predates the #3238 branch and is not caused by it.

## Open question for the fix

Whether the platform's auth service re-creates the missing service entries at startup was not determined. If it does, the defect is wasted work and a misleading auth.json; if it does not, the harness starts platforms with entries its own code believes it wrote.

## Acceptance criteria

1. Every entry the pre-seed adds is present in the instance's `auth.json` after `startup_platform` returns, asserted by reading the file back.
2. The fix does not rely on the caller remembering to build a new `AuthFile` per entry: either the write path reloads, or the pre-seed adds the entries in one operation.
3. The auth, vctl and testutils suites give the same results before and after.

Issue #3247, opened 2026-09-22
Test harness add_capabilities writes the grant to an entry it did not match


craigpnnl:
## Summary

`PlatformWrapper.add_capabilities` in `volttrontesting/utils/platformwrapper.py` selects the auth entry by public key and then writes it by user_id, so the grant can land on a different entry than the one it read.

At `develop` d68dff037b680383442576c634980a18a6aed8b7 (line numbers from that commit):

```python
    entry = auth.find_by_credentials(publickey)[0]     # :639, matches by key
    ...
    auth.add(entry, overwrite=True)                    # :647, writes by user_id
```

`AuthFile.add(overwrite=True)` resolves the write target through `_check_if_exists`, which stops at the first entry whose user_id matches. When two allow entries share a user_id and differ by key, the capabilities read from one entry are written over the other.

Two related defects in the same method:

- `entry.capabilities` is mutated in place. `AuthFile` holds a snapshot of the entries it read, so the mutation also changes that cached copy before the write.
- `find_by_credentials(publickey)[0]` raises `IndexError` when no entry matches, rather than reporting that the identity was not found.

This is test harness code. It was found while reviewing #3238, whose fix corrects the same pattern in a different method by writing at the matched index with `update_by_index`.

## Why it is separate from #3238

`add_capabilities` predates that change and grants to arbitrary identities rather than to the harness identity alone, and `.add_capabilities(` appears at many call sites across `volttrontesting/` and `services/`, so a behavior change there reaches well past that issue's acceptance criteria.

## Acceptance criteria

1. The capabilities are written to the entry that was matched, not to the first entry with the same user_id. A test seeds two allow entries that share a user_id and differ by key, grants to one, and asserts the other is unchanged.
2. The entry passed to the write is not an alias of the auth file's cached snapshot.
3. A public key with no matching entry produces a clear error naming the key or identity, not an `IndexError`.
4. Existing callers keep working: the suites that call `add_capabilities` give the same results before and after.

PR #3249, opened 2026-09-22
test: grant the harness dynamic_agent the control capabilities it uses

craigpnnl:
Fixes #3238.

The test harness identity `dynamic_agent` drives the platform lifecycle (`remove_all_agents`, `stop_platform`, `shutdown_platform`, and tests calling `prioritize_agent`), but held only `edit_config_store` and `allow_auth_modifications`. Since the control RPCs were gated, those calls are refused, which accounts for 31 of the 81 failures in a local run of the vctl and testutils areas, and probably the 6 that present as cross-test isolation failures.

This grants `dynamic_agent` the same five control capabilities the platform grants its own `control.connection` entry, named through the constants in `volttron.platform.agent.known_identities`: clear agent status, install and remove agents, start and stop agents, stop platform, tag agents.

Test harness only. No file under `volttron/` or `services/` changes.

## What the change does

- Grants the five capabilities on both grant paths: the startup pre-seed and `build_agent(identity="dynamic_agent")`.
- Updates a stale entry left by a prior grant or a reused `VOLTTRON_HOME`. The entry is matched by user_id and key, and written back at that index with `update_by_index`, so an entry with the same user_id under a different key is left alone.
- Reads and writes only through explicit paths under the instance's own `VOLTTRON_HOME`, so the update cannot reach another home even if a caller forgets the environment context manager.
- Assigns the exact capability set rather than merging, so a stale `driver_write` on a reused entry is removed.
- Skips the write when the entry already holds the exact set, and warns when no entry matches.

## Tests

15 tests cover the change, 5 of which start a real platform. Each fix is pinned by a test that was shown to fail against the code before it:

- the entry holds exactly the expected set, on both grant paths;
- an entry with the same user_id and a different key is untouched;
- a stale `driver_write` is dropped;
- no auth file or keystore is created outside the instance home, and the process environment is restored afterward;
- the grant does not run on an auth-disabled instance;
- a `start_stop_agents`-gated control call succeeds as `dynamic_agent` against a live platform, and fails with the platform's own refusal when the capability is removed;
- `build_agent` with no capabilities argument still gets its existing default;
- no write happens when nothing needs to change.

`volttrontesting/platform/auth_tests` is unchanged at 118 collected, 45 passed, 1 failed, 70 skipped, 2 xfailed, and the pre-existing failure is unrelated to this change. `tests/unit/test_rpc_allow_control.py` is 24 passed.

## Known limits

- Only `install_remove_agents` and `start_stop_agents` are exercised through a live platform. The other three are pinned by equality against the expected set; the harness does not call the methods they gate.
- A capability granted through a role or group would survive the exact-set assignment. Roles are not used in the harness outside the auth file's own unit tests.
- On a reused `VOLTTRON_HOME` whose `dynamic_agent` entry carries a different key, the grant does not happen and the warning is the only signal.

Related harness defects found during review and filed separately, not addressed here: #3247 and #3248.


craigpnnl (commented):
Review record, posted by the dispatcher from each reviewer's report.

Round 1, at 095e2de648813114f591f251a8b918e1a7ad3f4c:

| Lane | Verdict | Scope read |
|---|---|---|
| Code quality | APPROVE-WITH-NITS, 2 LOW | Both changed files whole, plus the diff |
| Security | REQUEST-CHANGES, 2 MEDIUM, 3 LOW | The diff whole, the changed regions plus `build_agent`, `with_os_environ`, `create_volttron_home`, `set_auth_dict`, `restart_platform`, and the auth file and keystore code the change calls |
| Test coverage | REQUEST-CHANGES, 1 HIGH, 2 MEDIUM, 2 LOW | Both changed files whole, the diff, the auth file exception hierarchy, and the control service gates; 3 mutants run |
| Error handling | APPROVE-WITH-NITS, 5 MEDIUM, 3 LOW | Both changed files whole, the diff, and the auth file, auth entry and keystore code the new path calls |

The deciding findings: the write-back could replace an auth entry that shared a user_id but not the key (found independently by two lanes, each reproducing it); the constraint that the harness must never touch a developer's `VOLTTRON_HOME` had no test, proven by a mutant that wrote to a second auth file and survived all six tests; and a stale `driver_write` capability survived the merge.

Round 2, at b9797efc5238d1c5bbb494312b64904c25951f19:

| Lane | Verdict | Scope read |
|---|---|---|
| Security | REQUEST-CHANGES, 1 MEDIUM, 2 LOW | The base-to-head diff, the changed method and its callers, and the auth file and keystore code; the round-1 tests re-run against the pre-fix code, 5 of 8 failing there |
| Test coverage | APPROVE-WITH-NITS, 2 MEDIUM, 3 LOW | Both changed files whole, the delta, the control service gates and the testutils workflow; 5 mutants run, each killing exactly one test |
| Error handling | APPROVE-WITH-NITS, 3 MEDIUM, 6 LOW | Both changed files whole, the delta, and the auth file, auth entry, keystore and home-resolution code |

All three lanes reported the same defect from different angles: the new test that proves the home-isolation property mutated `VOLTTRON_HOME` and deleted it afterward instead of restoring it, so later code in the same process resolved to the developer's real home.

Round 3, at 447c75a2c48be62754fab30de4b7e8e71da5356c:

| Lane | Verdict | Scope read |
|---|---|---|
| Security | APPROVE, 3 LOW | The round-2 delta whole, the changed method and its call site, the changed tests, and every `auth_enabled` site; the restore verified with the variable set, unset, and with an assertion failing inside the block |

Remaining LOW findings, recorded rather than fixed here: the warning defers to a later grant path that is itself silent, so a permanent failure to grant reads as transient; one test asserts the absence of a single exact phrase; and a capability granted through a role would survive the exact-set assignment, which is unreachable in the harness today.

Defects found during review that predate this branch are filed as #3247 and #3248. One more is not yet filed: the harness's environment context manager restores the Python-level environment but not the process environment, so a subprocess started afterward sees the older value.

Issue #3242, opened 2026-09-20
VIP RPC parameter-restriction regex silently loses full-string anchoring on `|` alternation (CWE-625), lets a restricted-capability agent escape the allowed-value scope


router0mail:
## Summary
VOLTTRON's VIP RPC subsystem lets a platform admin grant an agent a capability restricted to specific *parameter values* on the RPC method it gates (the "capability with parameter restriction" feature, `docs/source/platform-features/message-bus/vip/vip-authorization.rst`). The restriction value can be a `/regex/`-delimited pattern, e.g. `{"can_call_bar": {"x": "/.*"}}`.

The enforcement code builds the final regex by naively string-concatenating the admin's raw pattern between `^`/`$`:

`volttron/platform/vip/agent/subsystems/rpc.py:347-349`
```python
if _isregex(value):
    regex = re.compile("^" + value[1:-1] + "$")
    if not regex.match(args_dict[name]):
        raise jsonrpc.exception_from_json(jsonrpc.UNAUTHORIZED, ...)
```

If the admin's pattern contains a top-level `|` (alternation) -- the natural way to write "restrict this parameter to one of these several values", e.g. `/platform.driver|platform.actuator/` -- Python regex alternation has lower precedence than `^`/`$`. The compiled pattern is not `^(platform.driver|platform.actuator)$`, it's `(^platform.driver)|(platform.actuator$)`. `re.match()` only guarantees the match starts at index 0; it doesn't require the whole string to be consumed for the left-hand alternative (which keeps `^` but loses `$`), and for any alternative in the middle of a 3+-way alternation, **both** anchors are lost. Every alternative except the last degrades from "exact match" to "arbitrary-suffix prefix match".

## Reproduction (standalone, isolates the exact code path)
```python
import re
value = "platform.driver|platform.actuator"          # value[1:-1] of "/platform.driver|platform.actuator/"
regex = re.compile("^" + value + "$")                # rpc.py:348, verbatim construction

regex.match("platform.driver")                 # -> True  (intended)
regex.match("platform.actuator")               # -> True  (intended)
regex.match("platform.driverEVIL_OTHER_AGENT") # -> True  (BYPASS: not an allowed value, but matches)
regex.match("not_matching_at_all")             # -> False (correctly rejected)
```
3+-way alternation is worse -- every alternative except the last loses *both* anchors:
```python
regex = re.compile("^" + "aaa|bbb|ccc" + "$")
[...]

Issue #3237, opened 2026-09-17
Capability-gated RPC methods raise AttributeError when authentication is disabled


craigpnnl:
**Problem.** On a platform started with `allow-auth = False`, every exported method gated with `@RPC.allow` or `vip.rpc.allow` fails with `'Subsystems' object has no attribute 'auth'`.

**Cause.** `RPC._add_auth_check` (volttron/platform/vip/agent/subsystems/rpc.py) calls `self._owner.vip.auth.get_capabilities(user)` unconditionally. `Agent.Subsystems` creates `vip.auth` only when `enable_auth` is true.

**Affected methods:**
* the 7 config store edit methods, since commit 828636913
* the 11 ControlService methods gated in 41d24e23d and b83b8ea6b
* the 7 PlatformDriverAgent methods gated in f5156b7d5

As a result, on non-auth platforms `vctl install/start/stop/remove`, `vctl shutdown`, config edits and driver writes all fail.

**Proposed change.** In non-auth mode callers are not authenticated, so a capability check cannot identify them, and the documentation says the mode has no access restrictions. The fix makes the serving agent skip capability enforcement when it runs with authentication disabled, and leaves auth-enabled behavior unchanged. This restores access on non-auth platforms to what it was before 41d24e23d for control and driver methods, and opens config store edits, which currently fail.

**Acceptance criteria:**
1. **Auth-enabled behavior unchanged.**
   * All tests in tests/unit/test_rpc_allow_control.py pass unmodified.
   * A new integration test on the auth-enabled `volttron_instance` parameter: an agent built with default harness capabilities calls `control.prioritize_agent('foo', '100')` and gets a RemoteError containing `requires capabilities {'start_stop_agents'}`.
2. **Auth disabled reaches the method body.** On the auth-disabled parameter, the same call's RemoteError contains `Priority must be an integer from 0 - 99.`. It contains neither `requires capabilities` nor `has no attribute 'auth'`.
3. **Unit tests with a real object.** Use an owner whose `vip` object has no `auth` attribute (a plain object, not a MagicMock):
   * (a) Auth disabled: the method body runs once and its return value is passed through.
   * (b) Auth enabled: UNAUTHORIZED is raised and the body does not run.
   * Both tests fail at d68dff037 with AttributeError.
4. **Where the decision comes from.** The enforcement decision is taken from the serving agent's own authentication setting, captured at construction, never from the incoming message. If the setting is absent, enforcement is on.
[...]

Issue #3255, opened 2026-09-22
Integration coverage for capability enforcement with authentication disabled


craigpnnl:
## Summary

The fix for #3237 is covered by unit tests over the RPC subsystem: 54 tests exercise the capability gate, the flag capture, the startup warning and the parameter-restriction matching directly. What none of them show is the behavior the issue's acceptance criteria describe at the platform level:

- that `vctl install`, `start`, `stop` and `remove` reach the method body on an instance running with authentication disabled;
- that a config store edit succeeds there;
- that an auth-enabled instance still refuses a caller lacking the capability, end to end over the message bus rather than through a constructed subsystem;
- that each agent with gated exports emits exactly one startup warning on such an instance.

Two reviewers of that change said independently that the unit layer cannot establish these, and both recorded it as an open gap rather than a satisfied criterion.

## Why it was not done with the fix

The change is product code in `volttron/platform/vip/agent/subsystems/rpc.py`. A platform instance started by the test harness does not inherit the interpreter path of the test process, so it imports the platform from wherever it is installed rather than from the tree under test. An integration run in that setup exercises the installed copy, and a green result would say nothing about the change. Producing real evidence needs an environment where the platform subprocess loads the code under test.

## Acceptance criteria

1. An integration test starts an instance with authentication disabled and asserts that a capability-gated control method runs its body, asserting on the effect rather than on the absence of an error.
2. An integration test on an auth-enabled instance asserts that a caller without the capability is refused, and that the method body did not run.
3. The startup warning is asserted from the instance's own log.
4. The harness makes the platform subprocess import the tree under test, and the test proves it did, for example by asserting the module path the platform reports.
5. If criterion 4 cannot be met in this repository's harness, that is recorded here with what it would take, rather than the tests being written against an installed copy.

Issue #3254, opened 2026-09-22
RemoteError with a message and no exc_info raises UnboundLocalError


craigpnnl:
## Summary

`RemoteError.__init__` binds `msg` only inside `if exc_info:` (`volttron/platform/jsonrpc.py`), then passes it to `super().__init__(msg)` unconditionally. Constructing the exception with a message alone raises `UnboundLocalError` instead of a `RemoteError`.

```python
    def __init__(self, message, **exc_info):
        if exc_info:
            ...
                msg = message
            else:
                msg = '%s(%s)' % (exc_type, args)
        super(RemoteError, self).__init__(msg)
```

Latent in this repository: the producer always supplies `exc_type`, and there is one in-tree construction site. A third-party agent raising `RemoteError("some message")` hits it, and every `except RemoteError` handler misses the result, because what propagates is an `UnboundLocalError`.

Found during review of #3237 by a reviewer who hit it while building a probe. Read-verified at `develop` baca922d8489088ece5e99be3a0271d0185f1699.

## Acceptance criteria

1. `RemoteError("text")` constructs, and its message is that text.
2. The existing behavior with `exc_type` and `exc_args` is unchanged, asserted by a test.

PR #3256, opened 2026-09-22
fix: skip capability enforcement when authentication is disabled

craigpnnl:
Fixes #3237. Also fixes #3242, which is one line in the same function.

## #3237

With authentication disabled, every capability-gated RPC method raised `AttributeError` before its body ran: the check calls `self._owner.vip.auth.get_capabilities(user)` and the auth subsystem does not exist in that mode. That breaks `vctl install`, `start`, `stop` and `remove`, the config store edit methods and the driver write methods on a documented, supported mode.

This makes a gated method skip enforcement when the serving agent runs with authentication disabled, and log a warning naming the methods whose requirements are unenforced. Behavior on an auth-enabled platform is unchanged.

- The flag is read once from the serving agent's own core when the RPC subsystem is constructed, and never re-read. Nothing in an incoming message, the environment or a later mutation can change the decision.
- Only a real boolean `False` disables enforcement. A value that is not a boolean is reported in the log by type, not by value, and enforcement continues. An absent flag enforces.
- With authentication enabled and no auth subsystem attached, the call is refused as UNAUTHORIZED instead of raising `AttributeError`.
- Methods gated after construction, through the dynamic `allow()`, warn when they are wrapped.

Documented in `docs/source/platform-features/security/non-auth-mode.rst`.

## #3242

The parameter-restriction pattern was built by concatenating the operator's pattern between `^` and `$`, so a pattern holding a top-level alternation lost full-string anchoring: every alternative except the last matched as a prefix, and a restriction of `/platform.driver|platform.actuator/` accepted `platform.driverANYTHING`.

The match now uses `re.fullmatch`, which anchors each alternative on its own and also refuses a value that differs from an allowed one only by a trailing newline. Documented in `vip-authorization.rst`, beside the feature.

## What this widens, and what it does not

On an auth-disabled platform, the methods that raised `AttributeError` now run: 11 control methods, 7 driver write and override methods, the 7 config store edit methods, and any agent-defined gate. That is the documented behavior of a mode whose own documentation states it has no access restrictions, and it is how those methods behaved before the capability gates were added. An auth-enabled platform is unchanged, and the parameter-restriction change only ever refuses more than it did.

## Tests

`tests/unit/test_rpc_allow_control.py` goes from 24 tests to 54. The 24 existing tests are unmodified. Each new property was shown failing before the change that fixes it, and reviewers independently killed mutants against the added tests: hardcoding the flag, re-reading it at call time from the environment or from a live core, reporting a wrong count or a truncated method list in the warning, putting an address or a key in that warning, reverting the matcher, and accepting only the first alternative of an alternation.

## Known gaps

- No integration test covers the platform-level acceptance criteria: that `vctl` reaches a method body on an auth-disabled instance, that a config store edit succeeds there, and that an auth-enabled instance still refuses over the message bus. A platform subprocess does not import the tree under test in this repository's harness, so an integration run would exercise an installed copy and prove nothing about the change. Tracked in #3255, with what it would take.
- A separate route, an agent's own configuration file setting this flag, is tracked in #3250. It is unchanged by this pull request and no platform service uses that route today.
- Related defects found while reviewing this change, none of them touched here: #3251, #3252, #3253, #3254.


craigpnnl (commented):
Review record, posted by the dispatcher from each reviewer's report. Four lanes, three rounds, plus a confirmation pass.

Round 1, at 6aebfcb0dd11ec72027acb9ffbc339f68f2aa4b6:

| Lane | Verdict | Scope read |
|---|---|---|
| Code quality | APPROVE-WITH-NITS, 1 MEDIUM, 1 LOW | Both changed files whole, the diff, the issue and its comments |
| Security | REQUEST-CHANGES, 3 MEDIUM, 4 LOW | The diff whole, the changed method and its callers, the core, the agent constructor, the config-kwargs helper, the config store and auth subsystems |
| Test coverage | REQUEST-CHANGES, 3 HIGH, 4 MEDIUM, 1 LOW | Both changed files whole, the control service gates, the workflow; 10 mutants run |
| Error handling | REQUEST-CHANGES, 1 HIGH, 2 MEDIUM, 3 LOW | Both changed files whole, the auth file, auth entry, keystore and home-resolution code |

The deciding findings: only an absent flag failed closed, so a present-but-falsy value skipped enforcement while the string "False" enforced; a value differing from an allowed one by a trailing newline still passed the anchored pattern; a capability granted after construction skipped enforcement without appearing in the warning; and five mutants passed the whole suite, including one that hardcoded the flag and made the fix inert.

Round 2, at 24c32fa2237ef40c2d27e7ebe1b08fb7b2c5bc87:

| Lane | Verdict | Scope read |
|---|---|---|
| Security | REQUEST-CHANGES, 1 MEDIUM carried, 3 LOW new | The round delta whole, the changed method and call site, every flag site; 13 mutants run |
| Test coverage | APPROVE-WITH-NITS, 2 MEDIUM, 3 LOW | Both changed files whole, the delta; 18 mutants run, 15 killed |
| Error handling | APPROVE-WITH-NITS, 2 MEDIUM, 6 LOW | The changed file whole, the delta, the warning helper and both wrap paths |

The security lane's remaining item was that a separate route into the flag, an agent's own configuration file, needed an issue rather than a code change; it is #3250. The coverage lane found one branch of the new warning untested, and lowered one of its own earlier findings from HIGH to MEDIUM after establishing that the platform passes the flag explicitly at every service it constructs.

Round 3, at 1647378039c45c2087d40591e46cf412c289c4be: the remaining items closed, plus the documentation.

Confirmation pass, security lane, at 1647378039c45c2087d40591e46cf412c289c4be: APPROVE-WITH-NITS, 1 MEDIUM, 2 LOW, all in documentation. Equivalence of the simplified pattern construction was measured two ways, 900 pairs at the construction level and 361 through the real gate, with no case where the new head allows what the old one refused. With the flag enabled, the method body was reached on none of nine adversarial paths, with positive controls showing the probe could see a body run.

The documentation findings were fixed in d39e008edf61a0d511923f07f40dacbb08bf7ac9: the parameter-matching note moved to the page that documents the feature, since a parameter restriction is never evaluated when enforcement is skipped, and the claim about the warning was corrected to what the code does.

Not covered by any lane, and stated rather than implied: the platform-level acceptance criteria have no integration test, for the reason given in the pull request body. Tracked in #3255.

